PR #197, opened 2015-01-09
Removed unused coref code in NLP extractor

zifeishan:
Currently in nlp_extractor there is an additional step to extract
coreference which slows down the progress, but the coref data is unused
in the output table. Removed these steps to speed up the progress.

In the future, we should add coreference exraction into the NLP extractor.


rionda:
@zifeishan please don't just comment out code (especially without comments). Remove it if it's no longer needed.

zifeishan:
@rionda Sorry. These code to extract coref may be added into the nlp extractor in the future. Added a bit documentation to the commented code.

PR #199, opened 2015-01-10
Add support for automatically compute unit test coverage

rionda:
Add make target "coverage" and supporting code to compute the coverage of unit tests.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,20 +1,19 @@
 set -e
 
+# if greenplum supported (gpfdist exist), then test GP
+if hash gpfdist 2>/dev/null; then
+  echo "Testing Greenplum..."
+  bash test/test_gp.sh $1
 # if command psql exist, then test psql
-if hash psql 2>/dev/null; then
+elif hash psql 2>/dev/null; then
   echo "Testing psql..."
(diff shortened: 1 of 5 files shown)


rionda:
@zifeishan , if you have time can you please look at this?
Thank you.

test.sh
```diff
@@ -1,20 +1,19 @@
 set -e
 
+# if greenplum supported (gpfdist exist), then test GP
+if hash gpfdist 2>/dev/null; then
+  echo "Testing Greenplum..."
+  bash test/test_gp.sh $1
 # if command psql exist, then test psql
-if hash psql 2>/dev/null; then
+elif hash psql 2>/dev/null; then
```
zifeishan: @rionda It was intended to test Greenplum twice, one with "test_gp.sh" which tests with parallel grounding, the other uses non-parallel grounding. Maybe we could find a better structure for this?

rionda: @zifeishan : I see. I didn't get that. Ok, I'm going to revert it back to the original, but I'll add a comment explaining what you told me.

zifeishan:
Looks good to me. @rionda

PR #205, opened 2015-01-27
Refactoring grounding a bit

feiranwang:


src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -640,10 +569,42 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
+      CREATE TABLE ${VariablesObservationTable}(variable_id bigint primary key);
+      """)
+    calibrationSettings.observationQuery match {
+      case Some(query) => execute(query)
+      case None =>
+    }
+  }
+
+  // check whether greenplum is used
+  def isUsingGreenplum() : Boolean = {
```
zifeishan: Is it better to move this to Helper.scala? @feiranwang

zifeishan:
Looks good to me other than the comments. Have you tested on a GP machine? @feiranwang

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
feiranwang: I'll check that @zifeishan

feiranwang:
Not yet... will test it. Thanks! @zifeishan

chrismre:
Hi Guys, I wrote a code review that I shared with you via Google docs... this file needs a refactor. Let's get that done!

Issue #104, opened 2014-05-22
plpy_extractor run function cannot have 4-space indentation


zifeishan:
In run() function, if the first-level indent uses 4 spaces and there is "import_lib", everything will go into the "else" part in translated plpy function, and no errors will be reported.

OMG sorry for that.

We might remove all the first-level indent to NO indent in the function to solve it.

Also we should explicitly point out the translated files for easier debugging.

zifeishan:
Fixed in PR https://github.com/HazyResearch/deepdive/pull/251.

PR #209, opened 2015-01-30
Extractor parallel loading

zifeishan:
- Do not use environment variable `PARALLEL_LOADING` any more, use `extraction.parallel_loading` instead
- Updated the docs for parallel loading
- Updated BiasedCoin test sampler arguments for more stability

### DIFF
--- a/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
+++ b/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
@@ -30,7 +30,9 @@ import org.deepdive.helpers.Helpers.{Mysql, Psql}
  */
 object ExtractorRunner {
   
-  def props(dataStore: JsonExtractionDataStore, dbSettings: DbSettings) = Props(classOf[ExtractorRunner], dataStore, dbSettings)
+  def props(dataStore: JsonExtractionDataStore, dbSettings: DbSettings,
+      parallelLoading: Boolean = false) = Props(classOf[ExtractorRunner], 
+          dataStore, dbSettings, parallelLoading)
 
 
   // Messages
@@ -56,7 +58,8 @@ object ExtractorRunner {
 }
 
 /* Runs a single extrator by executing its before script, UDF, and after sript */
-class ExtractorRunner(dataStore: JsonExtractionDataStore, dbSettings: DbSettings) extends Actor 
+class ExtractorRunner(dataStore: JsonExtractionDataStore, dbSettings: DbSettings,
+    parallelLoading: Boolean = false) extends Actor 
   with ActorLogging with FSM[State, Data] {
 
   import ExtractorRunner._
@@ -368,11 +371,6 @@ class ExtractorRunner(dataStore: JsonExtractionDataStore, dbSettings: DbSettings
    */
   private def runTsvExtractor(task: ExtractionTask, dbSettings: DbSettings, taskSender: ActorRef) = {
   
-    // Determine if parallel loading and unloading is used by Env var
-    val parallelLoading = System.getenv("PARALLEL_LOADING") match {
-      case "true" => true
-      case _ => false
-    }
     log.debug(s"Parallel Loading: ${parallelLoading}")
     val loader = new DataLoader
     val udfCmd = task.extractor.udf
--- a/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
+++ b/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
@@ -398,7 +398,11 @@ class ExtractorRunner(dataStore: JsonExtractionDataStore, dbSettings: DbSettings
 
     // NEW: for mysqlimport compatibility, the file basename must be same as table name.
     val queryOutputFile = new File(queryOutputPath + s"${outputRel}.copy_query_${funcName}.tsv")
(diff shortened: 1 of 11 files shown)


zifeishan:
@zhangce @feiranwang  The env var was no longer used. Thanks!

zifeishan:
@chrismre The code itself works, but Sen proposed some changes to the configuration format that I haven't got time to add. 

I think we can first merge the current request, and I will update the configurations for parallel loading / grounding later. @SenWu do you think so?

chrismre:
Guys I'd like to get this into develop so we can have a few days with it! Can you fix this properly and then do the pull? It looks like a little bit of extra work to do it the right way :)

SenWu:
I don't the we can change the configuration file too frequently. We should remove parallel_grounding parameter and use only one parameter to control paralleLoading and parallelGrounding in dbsettings. @zifeishan

chrismre:
@zifeishan, I agree with @SenWu. If this change breaks the current application.conf format, we should make sure to talk about it as a group...

zifeishan:
@chrismre The current pull request is backward compatible. It just adds a configuration option `extraction.parallel_loading`. I agree that we should discuss before changing the format of `application.conf`, but the current pull request should be safe.

Sen was suggesting that we remove the `inference.parallel_grounding` and `extraction.parallel_loading` options, and I think this might need more discussion after merging this pull request.

chrismre:
@zifeishan If it's backward compatible, then no problem... move it in :)

chrismre:
@feiranwang @mikecafarella @zifeishan Is this going in?

src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
```diff
@@ -400,7 +398,7 @@ class ExtractorRunner(dataStore: JsonExtractionDataStore, dbSettings: DbSettings
 
     // NEW: for mysqlimport compatibility, the file basename must be same as table name.
     val queryOutputFile = new File(queryOutputPath + s"${outputRel}.copy_query_${funcName}.tsv")
-    val gpFileName = s"${outputRel}_unload_${funcName}"
+    val gpFileName = s"${outputRel}_query_unload"
```
mikecafarella: Zifei, I'm not sure what's happening on this line.  Can you weigh in?

zifeishan: I made this change because sometimes table names are too long (more than 64 chars) with the previous naming convention, and in this case the table will be incorrectly created with a wrong name, and some procedures in DeepDive will crash. Sorry for not commenting it in the code.

mikecafarella: So the loss of 'funcName' isn't crucial?  I am not sure where funcName comes from, so I can't tell if its loss will possibly cause a name clash here.

zifeishan: I just further updated the naming with `${funcName}_query_unload`, since it is unique and could prevent potential parallelism problems. I also added comments to this change. Waiting for the test.

chrismre:
@feiranwang @SenWu Does this interfere with your refactor or your coming revamp of loading? Is it easy for you to pull these changes? In particular, are we planning to get rid of the gpfdist stuff and replace it with the streaming interface?  

If this commit goes in, I would advocate that this go into the grounding refactor, not develop.

@zifeishan There are some things we should consider cleaning up before this pull goes in:

1.  Use optional instead of "" to represent nulls) as in https://github.com/HazyResearch/deepdive/pull/209/files#diff-0679872da5a85cac4903ab0bbf315847R25
I'm guessing this is probably a larger refactor of the way we parse the config file, which should be handled separately.

2. Who decide to deprecate the python in-database extractors, which are much faster than this approach? These kind of design decisions should be discussed, and I don't recall even a single email. Also, I don't agree with the statement or the decision right now. 

3. What does this portion of the commit have to do with parallel loading? https://github.com/HazyResearch/deepdive/pull/209/files#diff-f3eb1e9f4b57dbda44815a559475e4d9R64

feiranwang:
This doesn't have conflict with refactoring for now. We'll figure out how to deal with greenplum unloading when the streaming interface is ready to go.

zifeishan:
@chrismre Sorry for the delay.

For 1, the optional configuration handeling, I will leave it as it is, as it's a part of a larger refactoring.

For 2, regarding the python in-database extractors, we had a discussion in https://github.com/HazyResearch/deepdive/issues/104. There are some known unfixed bugs in plpy_extractor, and right now nobody is actively using this extractor type, so we once postponed the maintenance of that extractor type. I still agree that plpy_extractor is I/O-optimized, while tsv_extractor is computationally optimized, and there are cases that using plpy_extractor is faster. I will remove the deprecation in the documentation, and fix the bugs later.

For 3, sorry that this goes into the pull request. I will remove it.

mikecafarella:
@zifeishan It looks like a lot has changed here, and that you are planning to make further changes.  I will do a full review when your new changes are prepared.

zifeishan:
@mikecafarella I have done the changes mentioned in the last message. (For 2, I will not fix the plpy bugs for this pull request.) Thanks!

feiranwang:
Done in another pull request.

PR #224, opened 2015-02-06
removed relative setting of APP_HOME and DEEPDIVE_HOME in tutorial files

samzhang111:
The tutorial instructs the user to set $DEEPDIVE_HOME and $APP_HOME early on. It references these variables in the command line.

When run.sh and braindump.conf rewrite the environment variables, it uses relative paths, so forces user to run from a specific directory. Also the number of "../" is wrong for the app directory because it is set for the example directory, which has an extra nested level.

I think it makes more sense to ask the user to set these variables for themselves early on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ out/
 out/
 hsql:*
 *.pyc
+*.swp
 /lib/dw_mac
 /lib/dw_linux
 .cache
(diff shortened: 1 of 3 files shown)


netj:
Thanks for your patch.  However, we'll address this environment/relative-path-mess once and for all by introducing a nice facade command soon.  The extra level has been corrected now.

PR #230, opened 2015-02-10
Grounding refactoring

feiranwang:


chrismre:
@mikecafarella Mike, can you do a code review here? This is phase one of the grounding refactor. There is some more going on to remove obvious code redundancy. We want to make sure that all new code has comments, is factored to be easy to understand, etc. This push is not intended to address  redundancy issues (multiple data stores).

chrismre:
@feiranwang Ah good, I see now you did catch all the language drops.
@mikecafarella This should go out very soon. It corrects a major issue (CASCADE DROPS) and is generally cleaner code. Any chance we can get this out today? I'm even tempted to hotfix for 0.05 because the CASCADE DROP that Feiran removed are so awful....

mikecafarella:
@chrismre I am reviewing now...

src/main/scala/org/deepdive/datastore/Dataloader.scala
```diff
@@ -26,8 +26,10 @@ class DataLoader extends JdbcDataStore with Logging {
    * @param dbSettings: database settings (DD's class)
    * @param usingGreenplum: whether to use greenplum's gpunload
    * @param query: the query to be dumped
+   * @param folder: for greenplum, the relative path to gpfdist
    */
-  def unload(filename: String, filepath: String, dbSettings: DbSettings, usingGreenplum: Boolean, query: String) : Unit = {
+  def unload(filename: String, filepath: String, dbSettings: DbSettings, 
+    usingGreenplum: Boolean, query: String, folder: String) : Unit = {
```
mikecafarella: Is the intention to make "folder" a single-level directory, or can this be an arbitrary subdirectory structure?

feiranwang: It can be an arbitrary subdirectory structure. The folder argument is managed by DeepDive, and won't be exposed to users. If "folder" is invalid, "executeSqlQueries" will throw an exception.

src/main/scala/org/deepdive/inference/datastore/MysqlInferenceDataStore.scala
```diff
@@ -169,23 +169,6 @@ trait MysqlInferenceDataStoreComponent extends SQLInferenceDataStoreComponent {
         ORDER BY b1.bucket ASC;
       """
 
-    override def WRONGcreateCalibrationViewRealNumberSQL(name: String, bucketedView: String, columnName: String) = s"""
```
mikecafarella: I fully support removing methods that are prefixed with WRONG.

src/main/scala/org/deepdive/inference/datastore/PostgresInferenceDataStore.scala
```diff
@@ -104,5 +104,95 @@ trait PostgresInferenceDataStoreComponent extends SQLInferenceDataStoreComponent
...
+        count = rs.getLong(1)
+      }
+      return count
+    }
+    
+    // create fast sequence assign function for greenplum
+    def createAssignIdFunctionGreenplum() : Unit = {
+      if (!ds.isUsingGreenplum()) return
+
+      val sql = """
```
mikecafarella: This kind of construction makes me extremely nervous.  It's basically a huge amount of code that's entirely opaque to the build system; what if there's a stray keystroke here?  We won't know until something fails, weirdly.  

Can we move these function definitions into their own Scala wrapper objects, with corresponding isolated test code?

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -206,8 +156,7 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
     DROP TABLE IF EXISTS ${lastWeightsTable} CASCADE;
     CREATE TABLE ${lastWeightsTable} AS
       SELECT X.*, Y.weight
-      FROM ${WeightsTable} AS X INNER JOIN ${WeightResultTable} AS Y ON X.id = Y.id
-      ORDER BY id ASC;
```
mikecafarella: Why remove the ordering?  Was this an unnecessary element of the query?

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -345,88 +278,15 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
-    $$
-    LANGUAGE 'plpgsql';
-    """
-
   def init() : Unit = {
   }
 
+  // given a path, get file/folder name
+  // e.g., /some/path/to/folder -> folder
+  def getFileNameFromPath(path: String) : String = {
```
mikecafarella: How is calling "getFileNameFromPath(p)" better than "new File(p).getName()"?  Just to be all Scala-like?

src/main/scala/org/deepdive/inference/datastore/PostgresInferenceDataStore.scala
feiranwang: Sure. Will do.

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -459,173 +319,89 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
 
+    // variable observation table
+    ds.dropAndCreateTable(VariablesObservationTable, "variable_id bigint primary key")
+    calibrationSettings.observationQuery match {
+      case Some(query) => {
+        log.info("Executing user supplied observation query")  
+        execute(query)
+      }
+      case None => {
+        log.info("There is no o query")
```
mikecafarella: I am not entirely sure what an observation query is, so I don't know if failing to run it is catastrophic or not.  Is this for debugging, or for indicating the evidence variables?

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -459,173 +319,89 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
-        case RealNumberType => "('00001')"
-        case RealArrayType(x) => "('00001')"
       }
-      val cardinalityTableName = s"${relation}_${column}_cardinality"
+      val cardinalityTableName = InferenceNamespace.getCardinalityTableName(relation, column)
+      ds.dropAndCreateTable(cardinalityTableName, "cardinality text")
       execute(s"""
-        DROP TABLE IF EXISTS ${cardinalityTableName} CASCADE;
-        CREATE TABLE ${cardinalityTableName}(cardinality text);
         INSERT INTO ${cardinalityTableName} VALUES ${cardinalityValues};
```
mikecafarella: We might want to promulgate a new coding standard: no direct SQL queries without a function wrapper, into which parameters are passed.  The goal in doing that would be to enforce type checking by the compiler.  As it stands, if there's a type clash you don't find out until you try to run the query.  Not a huge deal for this commit, but probably a good idea in the future.

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -459,173 +319,89 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
+  // ground variables
+  def groundVariables(schema: Map[String, _ <: VariableDataType], du: DataLoader, 
+      dbSettings: DbSettings, parallelGrounding: Boolean, groundingPath: String) {
+        schema.foreach { case(variable, dataType) =>
+      val Array(relation, column) = variable.split('.')
+      
+      val variableDataType = InferenceNamespace.getVariableDataTypeId(dataType)
+
+      val cardinality = dataType match {
+        case BooleanType => 2
```
mikecafarella: Can we use a reserved word for the BooleanType constant here?  In another part of the code, the magic number is "0" instead of "2" and it is not clear whether that's a bug or not.

src/main/scala/org/deepdive/inference/datastore/InferenceNamespace.scala
```diff
@@ -0,0 +1,55 @@
...
+  def getVariableFileName(relation: String) = s"dd_variables_${relation}"
+  def getFactorFileName(name: String) = s"dd_factors_${name}_out"
+  def getWeightFileName = s"dd_weights"
+  def getFactorMetaFileName = s"dd_factormeta"
+  def getBackupFolderName = s"_dd_backup"
+
+  // variable data type id
+  def getVariableDataTypeId(variable: VariableDataType) : Int = {
+    variable match {
+      case BooleanType => 0
```
mikecafarella: This is what I'm talking about, regarding BooleanType reserved word.  See the "2" comment elsewhere

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
feiranwang: I think it's the same... just to wrap it up as a helper function and give it a better name...

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -656,41 +432,102 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
+    }
+  }
+
+  // convert grounding file format to be compatible with sampler
+  // for more information about format, please refer to deepdive's website
+  def convertGroundingFormat(groundingPath: String) {
+    log.info("Converting grounding file format...")
+    val ossuffix = if (System.getProperty("os.name").startsWith("Linux")) "linux" else "mac"
+    // TODO: this python script is dangerous and ugly. It changes too many states!
+    val cmd = s"python ${Context.deepdiveHome}/util/tobinary.py ${groundingPath} " +
```
mikecafarella: Use Java File object to build paths like this, not manual "/" chars and hard-coded strings.  File will take care of directory separators and control characters, etc.  Also, "util" and "tobinary.py" should be reserved constants

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -656,41 +432,102 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
+
+  def groundFactorsAndWeights(factorDescs: Seq[FactorDesc],
+    calibrationSettings: CalibrationSettings, du: DataLoader,
+    dbSettings: DbSettings, groundingPath: String, parallelGrounding: Boolean,
+    skipLearning: Boolean, weightTable: String) {
+    val groundingDir = getFileNameFromPath(groundingPath)
+
+    // save last weights
+    if (skipLearning && weightTable.isEmpty()) {
+      execute(copyLastWeightsSQL)
```
mikecafarella: If this clause is executed, does that mean that you will copy the last weights, and then immediately drop the destination table?  Not clear what the desired behavior is here.

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -839,67 +660,53 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
         if (isFixed || weightlist == ""){
           execute(s"""DROP TABLE IF EXISTS ${weighttableForThisFactor} CASCADE;
```
mikecafarella: DROP TABLE is such a toxic operation that we should consider making it a method that one has to call in order to generate the SQL.  Inside that method, we would have operations that ensure we are only dropping tables inside the DeepDive namespace.  I am worried that a typo in a tablename somewhere yields a name clash with user tables, and then catastrophe

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
feiranwang: It is marking some variables as "observed" and fixed. Technically, these are not variables. The difference between observed and evidence is during learning, we will fit observed variables (not sampled in the sampler). This is a feature ported from an earlier engine called Tuffy...

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
feiranwang: Here it means the domain size (cardinality) of a boolean variable is 2. The enum types are defined in InferenceNamespace, where we define the variable type mapping: type -> enum integer.

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
feiranwang: Will fix this.

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
feiranwang: The desired behavior is before dropping the weight table, save the learned weights from last execution to another table, and use it to fill in the new weight table.

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
```diff
@@ -929,58 +729,111 @@ trait SQLInferenceDataStore extends InferenceDataStore with Logging {
...
+   */
+  def groundFactorGraph(schema: Map[String, _ <: VariableDataType], factorDescs: Seq[FactorDesc],
+    calibrationSettings: CalibrationSettings, skipLearning: Boolean, weightTable: String, 
+    dbSettings: DbSettings, parallelGrounding: Boolean) {
+
+    val du = new DataLoader
+    val groundingDir = getFileNameFromPath(Context.outputDir)
+    val groundingPath = parallelGrounding match {
+      case false => Context.outputDir 
+      case true => dbSettings.gppath + s"/${groundingDir}"
```
mikecafarella: Again, use File here.

src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala
feiranwang: I think it's unnecessary.

feiranwang:
@mikecafarella The pull request has been updated in response to the code review.

PR #232, opened 2015-02-11
Suppress GP output and clarify Docker setup messages

adamwgoldberg:
@mikecafarella


adamwgoldberg:
We also might want to merge this into master after just in order to get a green build again. It never updated the icon even though we retested the green commit we discussed yesterday.

PR #235, opened 2015-02-15
Datastore refactoring

feiranwang:


src/main/scala/org/deepdive/datastore/PostgresDataStore.scala
```diff
@@ -28,4 +29,138 @@ object PostgresDataStore extends JdbcDataStore with Logging {
...
     }
 
+  /**
+   * input: iterator (of what?)  
+   * 
+   * - Create a temp CSV file
+   * - run writeCopyData to write   
+   */
+  override def addBatch(result: Iterator[JsObject], outputRelation: String) : Unit = {
+    val file = File.createTempFile(s"deepdive_$outputRelation", ".csv")
```
mikecafarella: This temp file could be huge.  But File.createTempFile() will put something in /tmp, which is often on a small volume.  This code is thus extremely prone to failing for lack of disk space.

We should have our own method called DD.createTempFile(), which is configured to store in a ".tmp" subdirectory where the config file has told us there is a lot of space.

mikecafarella:
Feiran, I went over the PR.  Everything looks good except that File.createTmpFile() call.  Can you address this?

feiranwang:
Thanks! I'll look into that.

feiranwang:
@mikecafarella The tmp files are now put into deepdive's folder

chrismre:
@mikecafarella @feiranwang What's our status here? I'd like to pull this...

PR #247, opened 2015-04-02
Parallel loading

SenWu:


netj:
It seems you should merge develop again.  Otherwise, looks good to me.  Nice cleanup!

PR #248, opened 2015-04-02
allow 1e10 multinomial cardinality

feiranwang:


netj:
Can you briefly summarize how this cardinality change impacts the grounding and whether there's any interaction with the sample?  Does it only lift the supported upperbound for multinomial?  Then can't we take the max length of the cardinality among the multinomial variables?

feiranwang:
We use a fixed length encoding for multinomial cardinality. Previously it's 5 digits, which supports up to 1e5 cardinality. I extended to 10 digits. Multinomial may be refactored later, but this PR is just an extension for cardinality.

netj:
So there aren't any side effects such as the size of the grounded factor graph increasing because of this for existing apps?  1e10 practically sounds big enough, but I still think it'd be better to decide the necessary number of digits based on user's schema.

Issue #73, opened 2014-05-05
plpy


zhangce:
kbp10docs=# SELECT (__X.__COLUMN).doc_id, (__X.__COLUMN).mention_id, (__X.__COLUMN).sentence_id, (__X.__COLUMN).word, (__X.__COLUMN).type
kbp10docs-# FROM (SELECT func_ext_mention( doc_id, sentence_id, words, pos, ner, lemma, character_offset_begin, character_offset_end ) AS __COLUMN
kbp10docs(# from  sentence) as __X;
ERROR:  length of returned sequence did not match number of columns in row  (seg7 slice1 madmax4.stanford.edu:50007 pid=33719)
DETAIL:  
	 while creating return value
PL/Python function "func_ext_mention"

Does not throw errors now

zifeishan:
This bug is actually a more general non-termination bug when executing SQL queries by a file. Fixed in https://github.com/HazyResearch/deepdive/pull/251.

PR #251, opened 2015-04-07
Fixed the plpy indentation bug

zifeishan:
Fixed a bug: when using 4-space indent in plpy extractors, the iterations are messed up. #104 

Added some tests for the fix.

Fixed a non-termination bug when executing SQL queries by a file and there are errors in the middle, which caused #73 .

### DIFF
--- a/src/test/python/test_ddext/input-1.py
+++ b/src/test/python/test_ddext/input-1.py
@@ -1,0 +1,98 @@
+#! /usr/bin/env python
+#! /usr/bin/env python
+
+# Imports written in this area is useless, just for local debugging
+import ddext
+from ddext import SD
+import itertools
+import os
+from collections import defaultdict
+
+# input: sentences.id, p1.id, p1.text, p2.id, p2.text
+# output: has_spouse
+# returns: 
+
+def init():
+    # SD['json'] = __import__('json')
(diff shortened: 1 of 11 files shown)


src/test/python/test_ddext/input-1.py
```diff
@@ -0,0 +1,98 @@
+#! /usr/bin/env python
+#! /usr/bin/env python
```
netj: Why is this line repeated?

src/test/python/test_ddext/input-2.py
```diff
@@ -0,0 +1,98 @@
+#! /usr/bin/env python
+#! /usr/bin/env python
```
netj: Why is this line repeated as well?

netj:
Looks good overall.  Adding tests for plpy certainly makes it easier to understand how ddext works, although I think we should avoid rewriting user's code and eventually redesign the plpy extractor interface to be more compatible with tsv extractors.

Btw how are the new tests triggered from the test.sh?  Can you wire that up and handle the inline comments as well?

zifeishan:
@netj Fixed the duplicated lines.

The tests are now wired up with the main `test.sh` from DeepDive.

PR #263, opened 2015-04-25
Run ANALYZE before grounding

alldefector:
see
https://github.com/HazyResearch/deepdive/commit/5fa7493e0c1a73c144ff79f7
5719856d72f7f8cc

### DIFF
--- a/examples/tutorial_example/step2-generic-features/experiment-reports/v00001/dd-out
+++ b/examples/tutorial_example/step2-generic-features/experiment-reports/v00001/dd-out
@@ -1,1 +1,0 @@
-/Users/zifei/deepdive/examples/tutorial_example/step1-basic/../../../out/2014-12-22T113824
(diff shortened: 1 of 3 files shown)


netj:
ANALYZE doesn't work for mysql, see: 4af3a508ca08b0153560f1835d91af18b33b0135

alldefector:
Thanks, Jaeho. Looks like it's getting a bit complicated to backport... Unless develop are going to be merged within a couple days, could you or Feiran patch master with the 5fa7493 change? It'd make grounding 50X faster for current users.

chrismre:
The MYSQL port is broken, and it is likely to remain so for a while. I
would prefer to push this patch, since MySQL is neglected (and was never
recommended...)

netj:
I'll take care of this asap and backport to master.

feiranwang:
I think the commit below might be better. It handles psql/mysql issue https://github.com/HazyResearch/deepdive/commit/0253dd711e403d577814efcab791d92c0ff79577.
Should I put that in?

netj:
I thought I put that in, but maybe not.  You can try that again, but a cursory look at 0253dd7 makes me even more confused: it seems analyzeTable() is returning string, but how is it getting executed?  At least the one I put in called execute().

This combinatorial explosion over the control flow is really a mess and I think we should simply strip the current mysql code and redo it when necessary, possibly in a way that's well isolated from the postgres/greenplum implementation.

Issue #259, opened 2015-04-24
make test fails


nlothian:
Clean checkout of deepdive, doing `make test` fails.

```05:05:56 [sampler] INFO  /home/ubuntu/deepdive/util/sampler-dw-linux gibbs
05:05:56 [sampler] INFO  Executing: /home/ubuntu/deepdive/util/sampler-dw-linux gibbs -w /home/ubuntu/deepdive/out/test_coin/graph.weights -v /home/ubuntu/deepdive/out/test_coin/graph.variables -f /home/ubuntu/deepdive/out/test_coin/graph.factors -e /home/ubuntu/deepdive/out/test_coin/graph.edges -m /home/ubuntu/deepdive/out/test_coin/graph.meta -o /home/ubuntu/deepdive/out/test_coin -l 500 -i 500 -s 1 --alpha 0.1 --diminish 0.99 --quiet
05:05:56 [sampler] INFO
05:05:56 [sampler] INFO  #################MACHINE CONFIG#################
/usr/share/sbt-launcher-packaging/bin/sbt-launch-lib.bash: line 41: 28595 Killed                  "$@"
make: *** [test] Error 137```

Running the sampler on its own fails too:

ubuntu@deepdive:~/deepdive$ /home/ubuntu/deepdive/util/sampler-dw-linux gibbs -w /home/ubuntu/deepdive/out/test_coin/graph.weights -v /home/ubuntu/deepdive/out/test_coin/graph.variables -f /home/ubuntu/deepdive/out/test_coin/graph.factors -e /home/ubuntu/deepdive/out/test_coin/graph.edges -m /home/ubuntu/deepdive/out/test_coin/graph.meta -o /home/ubuntu/deepdive/out/test_coin -l 500 -i 500 -s 1 --alpha 0.1 --diminish 0.99 --quiet

\#################MACHINE CONFIG#################
\# # NUMA Node        : 1
\# # Thread/NUMA Node : 1
\################################################

\#################GIBBS SAMPLING#################
\# fg_file            : /home/ubuntu/deepdive/out/test_coin/graph.meta
\# edge_file          : /home/ubuntu/deepdive/out/test_coin/graph.edges
\# weight_file        : /home/ubuntu/deepdive/out/test_coin/graph.weights
\# variable_file      : /home/ubuntu/deepdive/out/test_coin/graph.variables
\# factor_file        : /home/ubuntu/deepdive/out/test_coin/graph.factors
[...]

PR #266, opened 2015-04-28
Adds a super simple dependency check script

netj:
to prevent issues like #259


feiranwang:
Looks good. Since master is always failing now... I'll merge it.

PR #274, opened 2015-04-29
Shorter name for unloading extractor queries

zifeishan:
Currently the extractor input query might have a too long name (>64 chars) which causes a bug in GP. Used a shorter name for unloading.

See https://github.com/HazyResearch/deepdive/commit/dfe749dc1b0ac36bf0087efaf502112dd1937bc8.

### DIFF
--- a/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
+++ b/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
@@ -396,7 +396,7 @@ class ExtractorRunner(dataStore: JdbcDataStore, dbSettings: DbSettings) extends 
 
     // NEW: for mysqlimport compatibility, the file basename must be same as table name.
     val queryOutputFile = new File(queryOutputPath + s"${outputRel}.copy_query_${funcName}.tsv")
-    val gpFileName = s"${outputRel}_unload_${funcName}"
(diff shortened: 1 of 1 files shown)


netj:
Looks okay, but what if the outputRel is longer than 50 chars?  Shouldn't we use a purely generated temporary table name?

zifeishan:
@netj We have many places in our system that tries to create tables by appending strings to existing tables, e.g.:

https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/datastore/Dataloader.scala#L48

https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/inference/datastore/SQLInferenceDataStore.scala#L893

I think our system is implicitly supposing users to use not-too-long table names (<30 characters should be fine, we haven't tested its limit there). However in this bug, the system is stricter since it used  `"${outputRel}_unload_${funcName}"` and `funcName` is `"func_${task.extractor.name}"` so a 20-char table (e.g. relation_mention_features) will easily break it. After the fix it is coherent with other places.

I agree that eventually the system should not pose any implicit restrictions to the length of table names, or we have to mention what the restrictions are.

netj:
Good point.  I created #276 as a TODO.

PR #281, opened 2015-05-12
Add support for different semantics

SenWu:


netj:
How does this work/pass tests without the new sampler binaries?

SenWu:
@feiranwang will prepare a sampler (fixed all known bugs) for release.

PR #286, opened 2015-05-16
Add COVERALLS test

SenWu:
Coveralls try to find source folder in src/main/scala/org folder, but our src folder is not there. So create a link.


test/test_coverage.sh
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/env bash
+
+# Create a link for the code to current dir
+ln -s src/main/scala/org ./
```
netj: Is this the only way to make `sbt coveralls` work?  They should have some configurable settings.  I guess you probably tried everything, so let's write a clear comment **why** we do this business, not what it's doing. These comments aren't useful because they're longer than the actual commands and only repeat what they do.  Also, you probably want to say something about the required sbt and scala versions, here or in the build files or somewhere in test scripts.

README.md
```diff
@@ -2,7 +2,8 @@
 
 Licensed under the Apache License, Version 2.0. http://www.apache.org/licenses/LICENSE-2.0.txt
 
-Tested with Travis CI.
+Tested with Travis CI. and COVERALLS
```
netj: All caps and the period looks a bit weird.  Can we just place the badges right under the heading without any text?  The text seems redundant.

test/test_coverage.sh
netj: And perhaps more importantly, a link to coveralls.io as a comment here maybe very helpful for those who aren't familiar with this stuff.

SenWu: I tried several ways, but this is only way to work I think. Yes, let me add some comments about versions.

netj:
Nice to see the coverage from coveralls.io!  Other than the comments above, looks good.

test/test_coverage.sh
netj: I meant you don't have to explain each line in English.  It's much more useful to write the non-obvious: why you do a ln and rm around sbt coveralls and what other dependencies are there.

Btw should the `sbt coverageAggregate` be moved here?  Right before it's submitted?  And the file name is a bit misleading because it's not collecting coverage but only submitting, right?

project/plugins.sbt
```diff
@@ -1,3 +1,9 @@
...
+
 addSbtPlugin("org.xerial.sbt" % "sbt-pack" % "0.5.1")  // for sbt-0.13.x or higher
 
-addSbtPlugin("de.johoop" % "jacoco4sbt" % "2.1.6")
\ No newline at end of file
+addSbtPlugin("de.johoop" % "jacoco4sbt" % "2.1.6")
+
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.1.0")
+
+addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.0.0")
```
netj: We should state the sbt/scala requirements right here.

README.md
SenWu: They use this name COVERALLS, but I removed it in later commit.

test/test_coverage.sh
SenWu: Although I think this is a bad idea, I change it in later commit. Think if you write a program and when you finish it you will collect the data. But if you it move here, it looks like we don't collect data. People can test coverage on his own machine, and they need covaregeAggregate.

netj: I see. Then, let's just rename the script to `submit_coverage.sh` or so to make it clear.

netj:
Looks good, merging.

netj:
Could you port this to develop as well in another PR soon?  Because that's where we mostly want to monitor the coverage.

PR #289, opened 2015-05-17
Port spouse example to Postgres-XL

alldefector:
Shared sequence seems to be a big bottleneck for Postgres-XL. Key change is a radical UDF used to assign bigint IDs (by creating a new table to replace the old one).


raphaelhoffmann:
This looks great. Adds new UDF for fast fast sequence assignment similar to the existing UDF for greenplum.

PR #293, opened 2015-05-20
add coverage test for develop branch

SenWu:


SenWu:
@coveralls If I understand coveralls correctly, it will find the source code ./ here, while our source code is not here. We did this https://github.com/HazyResearch/deepdive/blob/sen-coverage/test/submit_coverage.sh to let coveralls access the source code. Do you a better way to do this, e.g., a parameter to set path of source code?

test/submit_coverage.sh
```diff
@@ -0,0 +1,8 @@
+#! /usr/bin/env bash
+
+# Create a symbolic link at src/main/scala/org to let coverage test retrieve all source,
+# and remove it after test
+# Ref: https://github.com/scoverage/sbt-coveralls
+ln -s src/main/scala/org ./
+COVERALLS_REPO_TOKEN=QknRiHqsMIzaOEbmSYtikFuxuVWEiPAJe sbt coveralls
```
netj: Why is the token exposed here?  To my understanding, we don't need it when running through TravisCI.
Please remove the line, and reset the token since it's been compromised.

test/test_mysql.sh
```diff
@@ -62,6 +62,7 @@ sbt "test-only org.deepdive.test.integration.BrokenTest -- -oF" && \
 sbt "test-only org.deepdive.test.integration.BiasedCoin -- -oF" && \
 sbt "test-only org.deepdive.test.integration.MysqlSpouseExample -- -oF" && \
 sbt "test-only org.deepdive.test.integration.ChunkingApp -- -oF"
+sbt coverageAggregate
```
netj: Where are the `sbt coverage` lines for mysql?

netj:
Spotted a few errors.  Please fix them.

PR #295, opened 2015-05-24
start using old dark data stuff

chrismre:


chrismre:
@zhangce @nejt We're bringing back the dark data terminology... people seem to like it.

chrismre:
I already make deployed it.

PR #301, opened 2015-05-25
pgxl port

raphaelhoffmann:
Added documentation for running DeepDive with Postgres-XL, as well as some basic query refinements for XL installations.

### DIFF
--- a/doc/doc/advanced/pgxl.md
+++ b/doc/doc/advanced/pgxl.md
@@ -39,25 +39,28 @@ We assume that the user executing these commands has sudo rights.
 
 ### Setting OS Parameters
 
-Set the following parameters in the `/etc/sysctl.conf` file: 
+Set the following parameters in the `/etc/sysctl.d/50-pgxl.conf` file: 
 ```
+sudo tee /etc/sysctl.d/50-pgxl.conf <<EOF
 kernel.sem = 1000  32000  32  1000
-kernel.shmmax = 429496729600  #400GB
+# up to 400GB shared memory
+kernel.shmmax = 429496729600
+EOF
 ```
 
 After making these changes, run
 ```bash
-sudo sysctl -p /etc/sysctl.conf
+sudo sysctl --system
 ```
 
-Set the following parameter in the `/etc/ssh/ssh_config` file:
+Finally, adjust ssh settings in `/etc/ssh/ssh_config` and `/etc/ssh/sshd_config` files:
 ```
+sudo tee -a /etc/ssh/ssh_config <<EOF
 StrictHostKeyChecking no
-```
-
-Finally, set the following parameter in the `/etc/ssh/sshd_config` file:
-```
+EOF
+sudo tee -a /etc/ssh/sshd_config <<EOF
 MaxStartups 100
+EOF
 ```
 After that, run
 ```bash
@@ -69,7 +72,7 @@ sudo service ssh restart
 ```bash
 sudo apt-get update
 sudo apt-get -y install -y screen curl git rsync openssl locales openssh-server openssh-client
-sudo apt-get -y install -y gcc flex bison make cmake jade openjade docbook docbook-dsssl
+sudo apt-get -y install -y build-essential gcc flex bison make cmake jade openjade docbook docbook-dsssl
 sudo apt-get -y install zlib1g-dev libreadline6-dev python-dev libssl-dev
 sudo localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 ```
@@ -77,12 +80,15 @@ sudo localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-
 Now set `$TARGET_DIR` to the directory into which you would like to install XL.
 ```bash
 TARGET_DIR=/opt/pgxl
+sudo mkdir -p $TARGET_DIR
+sudo chown $USER $TARGET_DIR
 ```
 
-Then add the following line(s) to `/etc/ld.so.conf` (replace `$TARGET_DIR` with dir):
+Then create a file `/etc/ld.so.conf.d/pgxl.conf` for the `$TARGET_DIR`:
 ```bash
-/usr/local/lib
+sudo tee /etc/ld.so.conf.d/pgxl.conf <<EOF
 $TARGET_DIR/lib
+EOF
 ```
 
 The remaining parts of the installation do not require sudo rights. However, make sure
@@ -138,6 +144,7 @@ Now set `$DATA_DIR` to the directory, where the database files will be stored.
 **Be sure that you have write permission to this directory**.
 
 ```bash
+DATA_DIR=$TARGET_DIR/data
 mkdir -p $DATA_DIR
 ```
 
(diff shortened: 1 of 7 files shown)


chrismre:
This looks awesome! Has anyone run through the tutorial?

doc/doc/advanced/pgxl.md
```diff
@@ -0,0 +1,310 @@
...
+pg_buffercache pgcrypto pgxc_clean pgxc_ctl pgxc_ddl pgxc_monitor stormstats \
+tablefunc tsearch2 unaccent; do
+    cd $BUILD_DIR/postgres-xl/contrib/$pkg
+    make; make install
+done
+```
+
+### Set XL related session variables
+
+To set Greenplum related session variables, modify your
```
netj: "Greenplum" still remains here

doc/doc/advanced/pgxl.md
```diff
@@ -0,0 +1,310 @@
...
+`~/.bashrc` script and add the line:
+
+```bash
+export PATH=$TARGET_DIR/bin:\$PATH
```
netj: There shouldn't be a backslash before the last `$PATH`.  Surrounding quote may be safer.
```bash
export PATH="$TARGET_DIR/bin:$PATH"
```

doc/doc/advanced/pgxl.md
```diff
@@ -0,0 +1,310 @@
...
+DATA_NODE_WAL_BUFFERS="16MB"
+DATA_NODE_CHECKPOINT_SEGMENTS="256"
+```
+
+Create a file `$TARGET_DIR/conf/pgxc_ctl.conf` with the following contents (replace
+variables with their values):
+
+```
+pgxcOwner=$USER
+pgxcUser=\$pgxcOwner
```
netj: There seems many unnecessary backslashes here for pgxc_ctl.conf.  Is this intentional?

I'd argue most steps involving config file creation should be written in the following style:

1. first ask the user to set several shell variables, 
2. then run a huge `cat` command with heredoc that produces the file.

It's easy for the reader since they can blindly copy/paste the whole thing into their terminal to just get things work and also clear how advanced readers could customize if they wanted.

For example:
```bash
# some customizable variables
USER=mylogin
pgxcOwner=mylogin
# more variables ...

# create pgxc_ctl.conf
cat >"$TARGET_DIR"/conf/pgxc_ctl.conf <<EOF
pgxcOwner=$USER
pgxcUser=$pgxcOwner
...
EOF
```

src/main/scala/org/deepdive/datastore/JdbcDataStore.scala
```diff
@@ -232,14 +232,16 @@ trait JdbcDataStore extends Logging {
   }
 
   // check whether postgres-xl is used
-  def isUsingPostgresXL() : Boolean = {
+  lazy val isUsingPostgresXL : Boolean = {
```
netj: `lazy val` seems good for these checks!

doc/doc/advanced/pgxl.md
```diff
@@ -0,0 +1,310 @@
...
+### Configure PATH
+
+Configure the following paths into your `~/.bashrc`:
+
+```bash
+export PATH=$TARGET_DIR/bin:$PATH
+export PGXC_CTL_HOME=$TARGET_DIR/conf
+```
+
+Run `source ~/.bashrc`.
```
netj: This section is repeated (L113: Set XL related session variable).

netj:
Looks great!  I read through it without executing any commands, so there may be some detail I missed, but anyone who's familiar with PGXL should be able to get DD up and running.  Maybe adding a link to the XL spouse_example may be also helpful?

raphaelhoffmann:
Thanks, Jaeho! I fixed the writing in response to your feedback. Will walk through the tutorial on a fresh ec-2 instance and see if copy-paste gets me all the way through. Will let you know.

netj:
@raphaelhoffmann Any updates?  Otherwise, I'd like to give a try.

raphaelhoffmann:
@netj It would be great if you could give it a try. Thanks!

netj:
I tested on a 14.04 ec2 instance, and the doc seems accurate after some minor edits.  Will merge.

netj:
After installing, I tried running `make test` and found a test isn't passing with Postgres-XL (PostgresInferenceRunnerSpec: Postgres inference data store, grounding the factor graph with Multinomial variables, should work with weight variables) with the following error:
```
org.postgresql.util.PSQLException: ERROR: Postgres-XL does not currently support ORDER BY in subqueries
[...]
     at org.deepdive.datastore.JdbcDataStore$$anonfun$executeSqlQueries$2.apply(JdbcDataStore.scala:71)
     at org.deepdive.datastore.JdbcDataStore$$anonfun$executeSqlQueries$2.apply(JdbcDataStore.scala:68)
[...]
```

Any ideas?

raphaelhoffmann:
Oh. I guess never tested multinomial variables. We'll have to rewrite that
SQL query for XL to avoid ORDER BY in the subqueries. I'll create a fix.

raphaelhoffmann:
@netj @zifeishan I removed two ORDER BY clauses that are used on temporary multinomial tables.
All tests are passing now on XL. However, I don't completely understand why we had the ORDER BY clauses there: Does one of you know? Are they OK to be removed? If not, there might be another way to make it work by changing the XL distribution strategy for these tables, e.g. setting it to replication.

https://github.com/HazyResearch/deepdive/blob/raphael-pgxl/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L601-L621

netj:
@feiranwang Feiran, could you comment on why we need those ORDER BY clauses for multinomial?  The original code had ORDER BY even for the _unsorted one with a comment about Greenplum.

feiranwang:
@netj I'm not sure why we need those two tables. There are comments by Ce on https://github.com/HazyResearch/deepdive/blob/develop/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L526. It seems to be used to fix some Greenplum issue according to the comment.

raphaelhoffmann:
@zhangce @netj @feiranwang Thanks, Feiran!
Ce, can you help us here?

netj:
If dealing with Greenplum was the sole purpose, this will be another good example why we should completely isolate each port at the cost of duplicating some SQL code.

netj:
@feiranwang Does the factors with "MultinomialFactorFunction" need to be grounded in a particular order?  I can hardly imagine that being the case, but you're the expert here.  These ORDER BYs actually originate from [your commit for multinomial support](https://github.com/HazyResearch/deepdive/commit/6919a5be0faa5845d1df5536989b7d1749d749a0#diff-d6b360f7700b78f35e3e5044dd377354R640) and it seems @zhangce just tried to fix some issues on GP.

feiranwang:
@netj I thought you were discussing the greenplum issue... Yes, MultinomialFactorFunction is grounded in a particular order for convenient array-like access to weights in the sampler. (Information about multinomial here http://deepdive.stanford.edu/doc/basics/schema.html, and https://github.com/HazyResearch/sampler/blob/master/src/dstruct/factor_graph/factor_graph.cpp#L51). I'll add comments for this.

feiranwang:
Added comments in #314

netj:
Thanks for your clarification, but what we need here is not adding the comments about current assumption. We're trying to eliminate the ORDER BY in the DeepDive grounding. The code you pointed seems to be dealing with compact factor after loading the input. So, is there any way we can do the ordering from the sampler side, since we're already doing lots of reordering and compaction there anyway?

raphaelhoffmann:
It should be possible to fix this in the database. Changing replication strategy doesn't help (XL still throwing error), but I think that we we might be able to fix this by changing our UDF for sequence assignment. It currently uses row_number(); we might be able to get the ORDER BY in here.

feiranwang:
@netj Briefly discussed with Raphael, this order by is used to make sure weight ids for multinomial factor are in a particular order. Raphael mentioned this may be achieved by changing the sequence assignment.

netj:
We decided to put this in without multinomial support for the moment.  On grounding multinomial factors, we discussed two possible solutions:

1. find a general solution for parallel/distributed databases to produce the correct order assumed by the sampler, e.g., using `row_number()`, or

2. produce extra info that can be easily generated by parallel/distributed databases and have the sampler recover the assumed order of weight ids.

Issue #200, opened 2015-01-11
factor_graph.pdf has dead link


adamwgoldberg:
http://deepdive.stanford.edu/assets/factor_graph.pdf links to http://deepdive.stanford.edu/doc/inference_rule_functions.html, which 404s.

netj:
@feiranwang Can you fix the pdf? The right link seems to be http://deepdive.stanford.edu/doc/basics/inference_rule_functions.html with `basics/` in between.

feiranwang:
The link seems to be fixed already. It links to a crappy tutorial I wrote...

netj:
No, please fix the broken link to inference_rule_functions.html in the second page of the pdf.

feiranwang:
ok I see

PR #305, opened 2015-05-26
fix dead link in factor_graph.pdf (#200)

feiranwang:


netj:
I'm thinking we could keep the source (.tex?) for the file together if that's not too much work.

feiranwang:
@netj Added source

netj:
Looks great, merging.

Issue #258, opened 2015-04-21
Inference table does not contain evidence variables


zifeishan:
In the sampler, the variables are divided into query (without supervision label) and evidence (with supervision label). Currently the sampler does not compute probability for evidence variables. (That's why #vars in the inference table does not equal to #variables in total.)

However In the context of distant supervision, users usually need to know the probability even for evidence variables, because distantly supervised labels are noisy. A workaround is duplicating the evidence variables and make the duplicates query, but it's not very elegant, may have problems in a complex factor graph, and it's currently mentioned nowhere in the docs.

I think the sampler should support computing the probability for evidence variables. It probably should be made default, and developers can specify `sampler_args` to turn it off. If we do not want the change, we should at least make developers more aware of it, and point them the right ways to handle it.

netj:
@feiranwang I thought this was now possible through a sampler option, no?  Maybe I'm confusing with #271?

netj:
It seems sampler's sample_evidence branch has the code: https://github.com/HazyResearch/sampler/commit/b721e65060a50f9d71500c7d2ff118e8d90bfa01#diff-bcc8c161c87b19685947789ac42e9fe9R32 @feiranwang Could you bring the binary here, also documenting the sampler option?

feiranwang:
It's different from #271 as this is for inference results. This feature is added in sample_evidence branch, I'll add it to release.

Issue #238, opened 2015-02-27
Imply factor function


zifeishan:
When is it desired to use a "Imply" factor function?

For my application, I want to punish the case where variable `a` is True and `b` is False. I tried `Imply(a, b)` but got the learned weights dramatically large.
I simply change it to `And(a, !b)` which should be the negation of the `Imply` factor, and it works properly (learns a small negative weight).

Anyone got similar experience and could explain this?

feiranwang:
There's a bug in imply function in sampler. I think Ce fixed it. Please check the compiled binary (release/dw_linux) under HazyResearch/sampler.

netj:
@feiranwang Wasn't this also fixed somewhere in sampler?  Could you get that merged with the code for #258 as well as any other sampler related ones, then bring the final binary in?

feiranwang:
Sure. That's what I'm thinking. I'll prepare samplers that include all these fixes for release.

PR #306, opened 2015-05-26
Update samplers

feiranwang:
As discussed in #238, #258.


netj:
Could you squash the submodule and binary update into a single commit, with a more descriptive message (what it fixes, why it had to be updated, etc.)?  Maybe even with the doc updates.  That way it's much clearer what those opaque changes are.

netj:
Thanks for squashing.  Should we turn the --sample_evidence flag on by default as suggested?  Is there a test that checks whether the flag works?

feiranwang:
I think it may be better to keep the default behavior as before and give users the option to change, for backward compatibility.
This arg was tested locally. I'll also add a test for this in sampler's tests.

feiranwang:
Test for sample_evidence was added in sampler's develop branch. I utilized the existing test, and add --sample_evidence to it, and checked the results.

Issue #277, opened 2015-04-30
Cannot have multiple variable columns in a same table


zifeishan:
Currently DeepDive cannot allow multiple variable columns in one table, since different variables need to have different `id`s and each table only has one ID column. However this is not mentioned in the doc, and we even have examples that are wrong:

http://deepdive.stanford.edu/doc/basics/schema.html

http://deepdive.stanford.edu/doc/basics/configuration.html#inference_schema

We should at least fix the docs.

chrismre:
@zhangce @zifeishan @netj 

How is this possible? When did this happen?

zifeishan:
@chrismre As I understand, this happens a long time ago, when we decided to do grounding by assigning an `id` column to each row in the variable table. Sampler would require different variables to have different `id`s, but it is not possible under the current grounding framework since each table that contain variables only have one `id` column.

chrismre:
Sure, but how did the docs get so far out of sync?!?

zifeishan:
I will fix the docs tonight.

chrismre:
@zifeishan let's wait for @netj to weigh in on the fixes. I'd like it if all the tutorials were in the examples directory and part of the tests.

@feiranwang This should also have some learning tutorial with simple results :)

zhangce:
>>  As I understand, this happens a
> long time ago, when we decided to do grounding by assigning an id column
> to each row in the variable table. Sampler would require different
> variables to have different ids, but it is not possible under the current
> grounding framework since each table that contain variables only have one
> id column.

I think we did not decide to add this constraint ("One variable columns in
a same table") due to whether our "grounding framework" can support it or
not :) -- If this constraint does not make sense, we should not alter the
semantic
just because it is easy to write grounding or the sampler...

I think the reason we made this decision is that we thought adding this
constraints will make it clear that each tuple in the database corresponds
to
one random variable. We thought this makes the semantic more clear to
the user; in the mean time, does not hurt our expressive power.

Ce

feiranwang:
Sure, I'll check with Zifei.

netj:
I like Ce's reason behind the decision. This was a well known limitation among us but the docs weren't updated timely. Let's fix the docs asap and clearly state why we restrict to one variable per relation.

[...]

PR #308, opened 2015-05-26
Fixed smoke examples in docs

zifeishan:
Fixed issue #277 . Made the docs coherent with the system when mentioning the spouse example.


zifeishan:
@netj Could you remove these changes? Thanks!

Issue #304, opened 2015-05-26
Shippable error


netj:
Shippable checks for every pull request has been marked as failure since a long time ago.  @adamwgoldberg Can we just disable this, or is it anyhow related to shipping Docker containers?  I checked the cause (which was non-trivial, not just a click), and it seems some `gem install` command in our travis.yml needs `sudo`, which isn't necessary for TravisCI.  If Shippable is just running the same tests as Travis, can we get rid of it?  Or will prefixing sudo to the gem command fix everything?

netj:
@adamwgoldberg It seems your shippable-branch contains a working shippable.yml.  However, I'm questioning why we need Shippable in addition to Travis.  Please comment on why we need Shippable if there's a straightforward reason. (maybe in connection to Docker?)

One nice part I see is Shippable gives a good summary on failed tests/assertions in Scala code.  However, if it doesn't provide any real extra value, then I think we should get rid of it to reduce maintenance burden.

netj:
Now, #311 will clear the Shippable error on our side, but tests using `COPY FROM` keeps failing due to their limitation (Shippable/support#1503).

adamwgoldberg:
Right, the original reason was to have performance testing on known hardware. I think that's something we can table for now. I'll remove it now.

netj:
I see.  Perf testing sounds like a compelling reason to keep container-based tests.  But for now, yeah let's turn it off if you can.

adamwgoldberg:
Turned it off. Opened #313 as well.

netj:
Thanks.  Did you just remove repo hooks from GitHub, or something else from Shippable?  It wasn't very clear to me how to turn it off.

adamwgoldberg:
I disabled the project on shippable.com and removed the repo hooks. I
suppose you didn't have access to the shippable login. Whoops. :)

netj:
That explains.  Thanks!

Issue #302, opened 2015-05-25
Unify automated installs


raphaelhoffmann:
We currently have automated installs in the form of 
* vanilla bash scripts
* Dockerfile scripts
* AMIs
* Virtualbox (for impala)

The problem is that it is hard to keep these up-to-date.

I suggest we drop support for Docker/AMIs/Virtualbox. Instead, we'll only have three bash install scripts:
1) install_with_pgxl.sh
2) install_with_pg.sh
3) install_with_greenplum.sh

Running one of these on an Ubuntu Docker container is easy. Running one of these on an Ubuntu VM on ec-2 is easy. Running one of these on an Ubuntu VM in Virtualbox is easy.

We can describe in one short paragraph how to run on any of these hosts.

netj:
This is a great idea.  I totally agree we should move towards this direction to reduce the maintenance burden.  I'll include new instructions in this release if you could share the scripts you have.

raphaelhoffmann:
Thanks – that's great. I just checked some scripts into branch
raphael-script.

You might want to adjust the scripts a bit.

The raw urls which you suggested work fine, eg.

curl -s
https://raw.githubusercontent.com/hazyresearch/deepdive/raphael-script/util/script/dduser_dd.sh
| bash

raphaelhoffmann:
@netj I cleaned up the scripts. The new installer works great.

Get an ec-2 instance with Ubuntu 14.04 and 100GB boot volume. Then run

curl -s https://raw.githubusercontent.com/hazyresearch/deepdive/raphael-script/util/script/entry.sh | bash

Could you link http://deepdive.stanford.edu/install to above script?

netj:
Very nice!  I added a temporary redirect to the website for /install.  It should be updated to point to the master branch upon release, but we can test things with your branch for now.  I think it's better to use raw.github.com instead and add `curl -fsSL`.

netj:
[...]

PR #311, opened 2015-05-30
Installer scripts with test cleanup

netj:
`util/install.sh` is the master installer script that detects operating system, loads platform-specific installers, and runs them in batch or interactive fashion.  Now, installing DeepDive can be done with a oneliner:
```bash
bash <(curl -fsSL deepdive.stanford.edu/install)
```

@raphaelhoffmann's scripts from #302 for installing DeepDive dependencies as well as Postgres-XL on Ubuntu has been slightly modified to fit into this modular, extensible structure.  A minimal script for Mac has been added mainly to show how the installer supports many platforms.  Each OS-specific install script can define as many installable components as it wants (by enumerating them from `list_installers`) and keep dirty details in separate files under install/ (loaded with `source_script` even over GitHub! See how install.Ubuntu.pgxl.sh is run).

Makefile and Travis config are rewritten to use the identical scripts, and this will unify the once diverging codebase towards local make, Travis, Docker, Shippable, AMI, etc.  As a side effect, tests were cleaned up, and broken link check running on Travis has been factored out as `make -C doc/ linkcheck` available to doc/ editors to check all external links as well in a minute or two.  Also, this cleanup is supposed to fix the always-broken-Shippable issue #304, but chunking test is not passing for some strange file permission error.


netj:
I'll update the installation docs in this PR.

netj:
On an Ubuntu instance, we can now test the scripts with:
```bash
export BRANCH=netj-installers
bash <(curl -fsSL https://raw.github.com/HazyResearch/deepdive/$BRANCH/util/install.sh)
```

raphaelhoffmann:
@netj This new set of installer scripts is fantastic!! Way better than I imagined! Testing it now.

netj:
@raphaelhoffmann Glad you like it!  I think we can polish scripts for postgres, XL, etc. over time, e.g., accepting important configuration variables interactively.

raphaelhoffmann:
Sounds good. I did some testing on an ec-2 ubuntu 14.04 instance. A few comments:

1. After one installation finishes, it simply asks "# Select what to install (enter a number or q to quit)?", but it doesn't show the options. Can we show the options again?

2. There are some dependencies between the different menu options.
- For example, running deepdive_git_repo without anything else results in an error due to missing git.
- Also, running deepdive_git_repo throws an error, if psql has not been installed yet:
lib/check-depends.sh: line 7: type: psql: not found
make: *** [depends] Error 1
## Failed installation for deepdive_git_repo

3. Trying the pgxl installer I got the following error
/etc/profile: line 4: PS1: unbound variable

4. For the pgxl installer, I need to be able to adjust the number of data nodes (and maybe target dir). The problem is that 16 nodes is only possible on machines with a very large amount of memory. Would it be possible to add some logic in there like I had in the install.sh in my branch?

5. Instead of setting up user/password for pg/pgxl, can we just do 
echo 'local   all   all   trust' | sudo tee -a /etc/postgresql/9.*/main/pg_hba.conf

netj:
Thanks for the comments!  I think for minor ones or ones with clear edits (6? and 7) you could modify the scripts on this branch directly from GitHub.  I found that feature pretty handy.

Taking care of dependency (2-4) is definitely something to figure out.  I'm tempted to simply support an optional precondition function (`can_install_*`) for every `install_*` functions, which controls visibility, e.g., show deepdive_git_repo only if git is there.  Regarding 4, psql probably shouldn't count as a requirement for deepdive.

1 and 5 are bash issues.  I'm relying on `select`, and that's the default behavior.  We could certainly put more effort polishing this part, but I wanted to keep the first version super simple with a bash builtin.  I have a nice set of bash snippets we can use later for creating slick interactive prompts.  5 is due to my `set -u` which prevents a lot of errors, but I guess we can turn that off by default for the rest of the installer scripts.  I'll quickly get this fixed.

raphaelhoffmann:
@netj Great. I think we're ready to merge.

raphaelhoffmann:
@netj I'd like to merge but I see that travis is failing because the spouse example gets an accuracy of 88% (not 90%). Is this something that can be ignored here? Let me know if I can go ahead and merge.

netj:
Thanks for your edits!  The tests are failing sometimes.  That needs to be fixed soon, but we know this PR is independent from the code that's failing.  So I think we can go ahead and merge.

raphaelhoffmann:
Thanks! This new installer will be a game-changer for DeepDive!

PR #315, opened 2015-06-03
handling assignIds with sort order

raphaelhoffmann:
Updated the XL sequence assignment UDF to take an optional ORDER BY clause.
Multinomials tests are now passing for XL.


netj:
My limited understanding keeps me from being perfectly convinced that this approach works exactly the same, but I'll trust you and @feiranwang on that.

PR #320, opened 2015-06-13
fix an incremental update statement for pgxl with distribution clause

raphaelhoffmann:


netj:
It seems the unit tests are now passing, but during the test_incremental, it still gives an error:
```
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: DROP SEQUENCE IF EXISTS dd_variable_sequence CASCADE                                                 
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: CREATE SEQUENCE dd_variable_sequence MINVALUE -1 START 0
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing SQL with callback...  SELECT num_variables FROM dd_incremental_meta_data
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: UPDATE dd_delta_has_spouse SET id = NULL
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: UPDATE dd_delta_has_spouse AS t0 SET id = t1.id 
        FROM has_spouse t1
        WHERE   t0.relation_id = t1.relation_id  AND (t0.label = t1.label 
        OR (t0.label is NULL AND t1.label is NULL))
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: DROP TABLE IF EXISTS dd_delta_has_spouse_inc CASCADE
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing SQL with callback... SELECT version() LIKE '%Postgres-XL%';
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: CREATE UNLOGGED TABLE dd_delta_has_spouse_inc AS  
        SELECT id, relation_id, label
        FROM dd_delta_has_spouse 
        WHERE id is NULL
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: ALTER SEQUENCE dd_variable_sequence RESTART 4685
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing SQL with callback... SELECT version() LIKE '%Greenplum%';
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing SQL with callback... SELECT version() LIKE '%Postgres-XL%';
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: SELECT copy_table_assign_ids_replace('public', 'dd_delta_has_spouse_inc', 'id', 4685)
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing SQL with callback... SELECT COUNT(*) FROM dd_delta_has_spouse_inc;
04:31:30 [PostgresDataStore(akka://deepdive)] DEBUG Executing query via JDBC: UPDATE dd_delta_has_spouse AS t0 SET id = t1.id 
        FROM dd_delta_has_spouse_inc t1
        WHERE   t0.relation_id = t1.relation_id  AND (t0.label = t1.label 
        OR (t0.label is NULL AND t1.label is NULL))
04:31:30 [PostgresDataStore(akka://deepdive)] ERROR org.postgresql.util.PSQLException: ERROR: could not plan this distributed update
  Detail: correlated UPDATE or updating distribution column currently not supported in Postgres-XL.
04:31:30 [profiler] DEBUG ending report_id=inference_grounding
04:31:30 [taskManager] INFO  Completed task_id=inference_grounding with Failure(org.postgresql.util.PSQLException: ERROR: could not plan this distributed update
  Detail: correlated UPDATE or updating distribution column currently not supported in Postgres-XL.)
04:31:30 [taskManager] ERROR task=inference_grounding Failed: org.postgresql.util.PSQLException: ERROR: could not plan this distributed update
  Detail: correlated UPDATE or updating distribution column currently not supported in Postgres-XL.
04:31:30 [taskManager] ERROR Forcing shutdown
```

netj:
@feiranwang This seems like the UPDATE statement that deduplicates by copying the base ids to the delta, right?  Is there any other way than updating the dd_delta table, such as doing an OUTER JOIN, by first using the base variable and COALESCE-ing to the id in delta or so?

feiranwang:
@netj Here we need to fill in id column (assign id) for variable table (dd_delta_has_spouse). Not sure I get your point... how to update a column without using UPDATE?

netj:
@feiranwang Sorry for not being clear.  I was thinking of an alternative approach creating a view to select the right id instead of overwriting a column.

feiranwang:
@netj I think we need those variable ids to be physically assigned in the variable table. Those variable ids are used in later parts, such as holdout, inference results, etc.

netj:
@feiranwang Yes, the ids will be physically assigned to both delta and base, but I think what you do there is you choose the ids in the base table if the delta variable already existed, right?  So all I'm saying is we could devise a nondestructive way to choose the id from the base for delta rows that join with a base row, and just fallback to the id assigned for delta otherwise.  I know this isn't a simple change, but I'm guessing we could've done this to avoid these compatibility issues with distributed databases.

feiranwang:
@netj I looked at the log again. It's actually id shouldn't be the first column (default distribution key). This is a documented requirement for variable tables in greenplum (http://deepdive.stanford.edu/doc/advanced/greenplum.html). This also holds for non-incremental version. I've updated develop_fix to reflect this. 

@raphaelhoffmann Sorry the problem that this PR is solving is actually not PXL's problem. I fixed the table schema to avoid updating a distribution key, so I think this PR doesn't need to go in.

raphaelhoffmann:
>
> @raphaelhoffmann <https://github.com/raphaelhoffmann> Sorry the problem
> that this PR is solving is actually not PXL's problem. I fixed the table
> schema to avoid updating a distribution key, so I think this PR doesn't
> need to go in.
>
> Sounds good. So, it seems both GP and XL use the first column as the
distribution key if no other column is explicitly set to be the
distribution key.

chrismre:
We should always set the partitioning key right?

On Sat, Jun 13, 2015 at 11:53 PM, Raphael Hoffmann <notifications@github.com

SenWu:
Right now, we just use the default distributed key setting (first column) in GP.

feiranwang:
For some temp tables, we use the default distribution key (the first column), and put a proper column as the first column. I think currently we don't write the distribution key explicitly due to two reasons:
1. we are sharing sql queries between different datastores
2. we don't know the exact schema of that relation, but we know which column should not be used as distribution key

chrismre:
Hmmm... this seems like a pretty important decision to be left to default.
Maybe we should revisit this later... Chris

feiranwang:
Yes, Jaeho proposed in the next refactoring to replicate all the sql queries for each datastore, so we can optimize each sql query for each datastore individually.

chrismre:
Jaeho is a wise man :)

netj:
Closing as we're solving this problem from #321.

PR #326, opened 2015-06-17
Piggy extractor (language-agnostic extractor driver with pipes)

alldefector:
Rebased on top of the 0.6 master.


alldefector:
Just rebased on master again.

doc/doc/basics/extractors.md
```diff
@@ -20,7 +20,12 @@ class contains different extractor *styles*:
   previous systems, but with limited performance
   - [`tsv_extractor`](#tsv_extractor): moderate flexibility and performance
   - [`plpy_extractor`](#plpy_extractor): parallel database-built-in extractors
-  with restricted flexibility
+  with seriously restricted flexibility because only the UDF function definition
+  is shipped to the DB
+  - [`piggy_extractor`](#piggy_extractor): a distributed language-agnostic pipe-based extractor driver.
+  We ship the entire runtime environment (code, data, and pip libraries) to
+  the DB server(s) and avoid multi-user collision by staging separate environments
+  for concurrent users. See `examples/spouse_example/piggy_extractor/`.
```
netj: Could you add more detail about using piggy?  There's no `#piggy_extractor` section defined here.  For example, what extra config is needed, how piggy knows what to ship, which databases are supported, etc.  It's probably a bad idea to have users stare at the piggy example and expect them to figure out what's changed from tsv or other examples.

examples/spouse_example/piggy_extractor/application.conf
```diff
@@ -0,0 +1,56 @@
...
+      udf_dir: ${APP_HOME}"/udfs"
+      udf: "python people.py"
+      dependencies: ["ext_clear_table"]
+      style: "piggy_extractor"
+    }
+
+    ext_spouses {
+      input: "people_mentions"
+      output_relation: "has_spouse(sentence_id, person1_id, person2_id)"
+      udf_dir: ${APP_HOME}"/udfs"
```
netj: It seems `udf_dir` is what's being shipped by piggy, right?  A comment around this line would be extremely helpful if so.

src/main/scala/org/deepdive/datastore/FileDataUtils.scala
```diff
@@ -17,4 +20,31 @@ object FileDataUtils extends Logging {
       reader.close()
     }
   }
+
+  def recursiveListFiles(dir: File): Array[File] = {
```
netj: Could you add unit tests for these extra functions?  The test coverage got decreased.

src/main/scala/org/deepdive/helpers/Helpers.scala
```diff
@@ -171,4 +172,23 @@ object Helpers extends Logging {
         s""" --silent -N -e '${query.replaceAll("'", "'\\\\''")}' """
     }
   }
+
+
+  def slugify(input: String): String = {
```
netj: Unit tests for these as well.  Not sure why this helper is needed from where.  Would be helpful if there's a brief comment.

util/piggy_prepare.py
```diff
@@ -0,0 +1,339 @@
...
+            import time
+            from datetime import datetime
+            from Queue import Queue, Empty, Full
+            from threading import Thread
+
+            # Create a pipe for logs.
+            # Only one process would succeed on the same host.
+            log_pipe_name = os.path.join('%(dir)s', '%(fname)s.pipe')
+            try:
+                os.mkfifo(log_pipe_name)
```
netj: I don't think it's a good idea to use named pipes for logging.  These log pipes will quickly become the bottleneck for any real app because all udfs will block on their `piggy.log()` calls until the `*_getlog` FUNCTION drains the pipe.  It seems the `GetLogThread` will only consume up to a megabyte from the pipe only once every second.  I think buffering in a temporary file is better in this case.  Also providing an easy way to toggle `piggy.log()` would be better than asking users to comment out all those calls.  Another way could be optionally starting GetLogThread based on DeepDive's log level and `piggy.log()` ignoring the passed lines if no thread will be reading, but this sounds too complicated already.

netj:
Please see my inline comments.  Otherwise, it looks pretty good except that most of the new parts aren't under the standard test suite.  It'd be nice to run the piggy_extractor example as another integration test if that's possible with a simple shell script similar to test/test_incremental.sh. I'm planning to rewrite all integration tests in Scala into simple shell scripts that directly test code under examples/, so let's defer this if you're tempted to copy/paste your example into src/test/scala/.

src/main/scala/org/deepdive/datastore/FileDataUtils.scala
alldefector: done

src/main/scala/org/deepdive/helpers/Helpers.scala
alldefector: It's used for use client hostname as part of the path of distributed package copies.

util/piggy_prepare.py
alldefector: Is there a pipe-like mechanism not subject to the small capacity limit that I could use? I used temp files earlier but it requires repeated scans (or complicated error-prone offset bookkeeping); so ideally I'd like to not go back to that approach. Any suggestions? Anything from unix that's like an append-only sorted string table?

alldefector:
Added documentation and unit tests. Will need to think about logging.

util/piggy_prepare.py
alldefector: Implemented the last option: https://github.com/HazyResearch/deepdive/commit/f6f6ef5f9525c8623e83e7f55c51c5a72e3db8ae

However, DD *always* gets complete log -- apparently due to some lack of flushing? Could you take a look and see what might be wrong? (I've tried various buffering options for both reader and writer of the named pipe.)

netj: I'm not sure I understand what you mean by log flushing.  This seems to be getting overcomplicated.  I think we shouldn't waste time reinventing a very limited remote shell as a Python hack.  It appears to me most of these issues would just go away with a simple `tail -F` over `ssh` to each host.  Can you make DD simply spawn a `for h in ${hosts}; do ssh $h tail -f ${log_pipe_name} & done; wait` command from GetLogThread and hang up all ssh processes once the thread has to die?

alldefector: Never mind. Found the issue -- DD was creating a new DB connection every time we call get_log, and the shared dict in play does not go beyond one process / connection, so the sucker thread became an orphan, and subsequently calls try to create more sucker threads.

Direct SSH access would definitely make things much easier, but the downside with it is that 1) the DB server / cluster doesn't always allows ssh access to all DD developers who'd like to use the DB; 2) we'd have to manage the SSH keys / passwords in addition to the JDBC params in the DD client. Longer term we may well have SSH access as a deployment requirement. I'd be as happy as you will be to switch to ssh when that day comes :)

netj: @raphaelhoffmann and I had a chat about the SSH requirement and we concluded it doesn't make sense to support the case where the database nodes disallow SSH access, because the DD user will most likely be the one who sets up the database as well. (We're making it easier so this is plausible.)  Therefore, the user has to have SSH access to even start the datanode on such host. (Sharing a parallel database between users never worked in practice anyway, let alone the case where a single user tries to run multiple queries..)

SSH authentication issues can always be handled transparently with ssh-agent or Kerberos.  Adding additional config params to DD will only add unnecessary complexity and I know such approach always break at some point.

Anyway, so is it possible to fix the issue with minimal change?

alldefector: Yes, it should be an easy fix.

I guess the fundamental question is what role the machine running DD plays (the master process orchestrating everything vs. a small client that should eventually run on a Chromebook). There is actually a third topology -- a monolithic one where one host runs the DB, the DD driver, extractions, and inference -- that's closest to the current status quo.

In the end, we'll likely have both the orchestrating master and the thin client. One way to move toward that direction is to start defining different services and clear interfaces between them. Those interfaces should probably be higher level than arbitrary local (psql) or remote commands (ssh) even if ssh is the simplest way to run things remotely. That said, streaming logs via SQL is clearly not a good interface either, though it does stick to the interface constraint that everything happens on JDBC (as opposed to having psql and ssh as well).

alldefector:
@netj yes, I'd leave the integration tests to you.. thanks!

I'm done with this prototype for now. Could you take another look at the new changes? I'm still using fifo with an eager log sucker thread and a 100K-line circular buffer in memory.

Issue #325, opened 2015-06-15
Improve Tests


netj:
* [x] Rewrite integration tests (out of Scala) in bash/bats to directly test the actual examples
    * [x] Bring [Bats](https://github.com/sstephenson/bats) in as submodule
* [x] Make sure tests can run with instrumented classes for test coverage instead of standalone jar
* [x] Make it easy to run different tests selectively without commenting out anything
* [x] Make it easy to switch backend database under test
    * [x] Standardize backend database setup scripts for every example
* [x] Exploit Travis build matrix to [speed up the build](http://docs.travis-ci.com/user/speeding-up-the-build/) and get results earlier (from currently ~20-30mins down to 5-10min?)
    * [x] Find a way to aggregate coverage across the build matrix for coveralls
* [x] Make sure no test configuration is hard coded and easy to override
* [x] Make sure tests can be performed by multiple users on a single host and shared database instance
* [ ] Have Travis perform tests with PGXL
* [ ] Have Travis perform tests with Greenplum

PR #328, opened 2015-06-19
whitespace cleanup + style check

netj:
After having trouble dealing with whitespace changes from the last release, I decided to add a script that enforces some very basic coding styles and cleanup all the previous mess:
- [no files without eol at the end](http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline)
- no DOS line endings (CRLF), always use Unix (LF)
- no trailing spaces
- no tabs in indentation (except Makefile, raw data)

Travis will always first check these styles and proceed to the rest of the build and tests.  Developers can run check locally before pushing:

```bash
make checkstyle
```

Let's rebase all our ongoing work on this PR and completely forget about whitespace issues from now on.


netj:
[Adding w=1 to the diff](https://github.com/HazyResearch/deepdive/pull/328/files?w=1) will show just the script and modification to Makefile and travis.yml.

badder:
This is great! Maybe also add some git hooks to enforce from the client side? Travis takes too long..

alldefector:
This is great! Maybe also add some git hooks to enforce from the client side? Travis takes too long..

E.g.: http://stackoverflow.com/questions/591923/make-git-automatically-remove-trailing-whitespace-before-committing

netj:
I'm reluctant to add it to developer's normal workflow.  Anyone who set up the editor appropriately will never create violations, so I think it's only going to slow down the innocent.  Gatekeeping bad code from Travis for PRs seems enough to me.  Speeding up Travis tests is another thing I'm planning to do in #325.

I made sure one can easily check what's wrong with the make command and open the problematic files from an editor if they ever need to:
```bash
vim $(make checkstyle)
```

I found an [interesting tool to deal with whitespaces from the Q&A](https://github.com/dlenski/wtf).  I guess we should rewrite the checks with it sometime later.  The script I think should eventually be rewritten as another test suite with a proper test framework like [Bats](https://github.com/sstephenson/bats) or so.

alldefector:
I don't really understand the script, but it looks quite legit! How do I approve the PR?

netj:
Thanks for reviewing. There's no formal button for approving PRs. You can just express approval like this and since we're both commiters either of us can press merge! :)

PR #331, opened 2015-06-24
Adds a proper `deepdive` command defining a standard app layout

### LINKED ISSUE #323: Provide a proper `deepdive` command that operates on a clearly defined standard layout for apps
* [x] Clean up sampler's `LD_LIBRARY_PATH` mess with a wrapper script #327
* [x] Define a standard layout for DeepDive apps:
    * application.conf
    * db.conf
    * udf/
* [x] Add a `deepdive` shell command that
    * sets `APP_HOME` by recognizing the top of a standard layout DeepDive app
    * sets `DEEPDIVE_HOME` to the source tree based on the path of itself
    * takes care of `PYTHONPATH` setup
* [x] Get rid of `sbt pack`
* [x] Update all examples to follow the standard layout
* [x] Update all examples and docs to use `deepdive` command instead of messing up environment from run.sh and calling `sbt`

netj:
See the new [DeepDive app structure and operations documentation](https://github.com/HazyResearch/deepdive/blob/netj-deepdive-facade-command/doc/doc/advanced/deepdiveapp.md#readme) for a quick summary of the changes.  The [rewritten walkthrough](https://github.com/HazyResearch/deepdive/blob/netj-deepdive-facade-command/doc/doc/basics/walkthrough/walkthrough.md#readme) may also be helpful for recognizing the changes from user's perspective.

The new `deepdive` *facade command* added by this PR basically encapsulates all sorts of DeepDive's internal complexity that used to be exposed to users over various shell scripts.  The standard app directory layout defined alongside the command establishes a clean interface between DeepDive infrastructure and the domain-specific application, which can also serve as a foundation for building other tools and languages.  Users no longer need to worry about where to put their stuff or invent new ways of writing env.sh, run.sh, etc., but they can now interact with DeepDive entirely through a single `deepdive` command with no confusion at all, just as we use the `git` command on any repo.

The command abstracts only a few of the most common operations users perform: initializing/setting up the database (`deepdive initdb`), running SQL queries over the database (`deepdive sql`), and running DeepDive extractors and inference (`deepdive run`).  Now each of these are all doable with a simple command across all apps, and the various pieces of ad-hoc shell scripts are wiped out or slimed down to a standard location.

Another significant change is the new output and logging location.  Upon every `deepdive run`, a new output directory is created under the app directory's `run/` and the log file is kept within it with name `log.txt`.  Also, the pipeline of the app to run can be specified over command-line argument, so no more nonsense of modifying a file every time to run different parts.

One last superficially significant change is that `application.conf` is now renamed to a more meaningful name `deepdive.conf`.  This inessential renaming is intentional to make it easy to tell whether an app is following the new style or not.  The command doesn't have a way to run application configuration file with a different name yet, but this should probably be added later to migrate the incremental example to the new style as well.

The entire facade command implementation is kept under `shell/`, and they are all short, straightforward shell scripts (~30 lines each).  Different database backends are supposed to provide a *driver* under `shell/driver.NAME` where the `NAME` is the URL scheme user writes in the application's `db.url` file.  Two drivers `postgresql` and `mysql` have been implemented and other variants `greenplum`, `postgresql-xl`, and `mysqlcluster` share them.  The application.conf's `db.default` now became mostly redundant and unnecessary, so at a later point should be removed.  The command was carefully written to make it extremely easy to add new subcommands by just adding an executable whose name starts with `deepdive-`.  This will enable us to quickly add useful features for interacting with the data products and analyzing them.

As this change affects so many documentation pages and example code, it had to touch almost all parts of the code base except the Scala code and the utilities.  Sorry for the huge diff, but this cleanup had to be done at some point!  Some documentation pages are still left as they were, only stating the page or section is outdated.  I need help on rewriting them from whoever gets excited by this new clean style. @SenWu @feiranwang @zifeishan @raphaelhoffmann @alldefector @ajratner volunteers?  I'm expecting various amounts of code review from everyone.

This closes #323.


alldefector:
Awesome change! Haven't looked at all the scripts. Would be nice if there is cmd option for the app directory and/or app conf file name.

netj:
@alldefector Using alternative paths is now possible by aebb987bf39eb766a3831b02b6d612f0c3f5f5e7

PR #333, opened 2015-07-11
Improves test a lot

netj:
Did most tasks in #325.  Builds on top of #331.

* Uses BATS to run ScalaTests as well as integration tests.
* Tests examples directly, instead of keeping redundant copies that's prone to getting out of sync.
* Most importantly, individual .bats file can be run very easily with `make test ONLY=...`.
* Now tests run on Travis 2-3x faster, from 20-30mins down to 10-15mins.
* Also, thanks to BATS and .travis.yml refactoring, test output now became very succinct, much much easier to notice what went wrong when a test case fails.
* Travis build matrix wasn't helpful since
    - we only have two parallel jobs, and
    - the test time was only half or third of the whole time, and
    - the parallel jobs don't always start at the same time.
* Tried container-based infrastructure with caching, and it seemed very fast and promissing, but failed due to lack of postgres-plpython.


chrismre:
This is awesome!

PR #342, opened 2015-07-25
Ddlog and incremental grounding update

feiranwang:
Update incremental grounding to match the weight column naming generated from ddlog since     HazyResearch/ddlog@a24788e27ac65d4118a518f1b92f9f2d950af02f.


chrismre:
@feiranwang @netj after code review, let's mail back the user who needed this!!

src/main/scala/org/deepdive/datastore/MysqlDataStore.scala
```diff
@@ -56,7 +56,7 @@ class MysqlDataStore extends JdbcDataStore with Logging {
       case "text" | "varchar" => s"convert(${expr.toString()}, char)"
       // in mysql, convert to unsigned guarantees bigint.
       // @see http://stackoverflow.com/questions/4660383/how-do-i-cast-a-type-to-a-bigint-in-mysql
-      case "bigint" | "int" => s"convert(${expr.toString()}, unsigned)"
+      case "bigint" | "int" => s"convert(${expr.toString()}, signed)"
```
netj: I'm unsure if this retains the bigint conversion effect.  Do the ids need be -1?

src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
```diff
@@ -547,11 +547,7 @@ trait SQLInferenceRunner extends InferenceRunner with Logging {
                 |= t1.${dataStore.quoteColumn(s"${InferenceNamespace.getBaseTableName(v.relation)}.id")}
                 |""".stripMargin.replaceAll("\n", " ")).mkString("AND")
             val weightJoinlist = factorDesc.weight.variables.map(v => {
-              // split column to get relation name
-              val colSplit = v.split('.')
-              val lastvrel = InferenceNamespace.getBaseTableName(colSplit(0))
-              val lastv = s"${lastvrel}.${colSplit.takeRight(colSplit.length-1).mkString(".")}"
-              s""" t0.${dataStore.quoteColumn(v)} = t1.${dataStore.quoteColumn(lastv)}"""
+              s""" t0.${dataStore.quoteColumn(v)} = t1.${dataStore.quoteColumn(v)}"""
```
netj: Is this backwards compatible?  Why did we need such complicated column name manipulation before if the same name could be just shared across t0 and t1?  Does this require any change to t0 or t1?

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
```diff
@@ -586,6 +586,120 @@ trait SQLInferenceRunnerSpec extends FunSpec with BeforeAndAfter { this: SQLInfe
...
         assert(numVariablesInc === 20)
       }
+
+      it("should work with materialization and incremental modes (multinomial)") {
+        cancelUnlessPostgres()
+        // XXX Incremental workflow not supported on Postgres-XL as inference.PostgresInferenceRunner.groundVariables raises the following error:
+        //   org.postgresql.util.PSQLException: ERROR: could not plan this distributed update
+        //   Detail: correlated UPDATE or updating distribution column currently not supported in Postgres-XL.
+        if (dataStoreHelper.isUsingPostgresXL) pending
+        // XXX Incremental workflow not supported on Greenplum as datastore.PostgresDataStore.createTableIfNotExists raising following error:
```
netj: I'm wondering if this is still the case.  Didn't we fix at least for GP?

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
```diff
@@ -586,6 +586,120 @@ trait SQLInferenceRunnerSpec extends FunSpec with BeforeAndAfter { this: SQLInfe
...
+          Map("id" -> i, "weight" -> s"weight_${i}", "is_correct" -> i%2)
+        }
+        dataStoreHelper.bulkInsert("r1", data.iterator)
+
+        val dbSettingsMat = DbSettings(dbSettings.driver, dbSettings.url, dbSettings.user,
+          dbSettings.password, dbSettings.dbname, dbSettings.host, dbSettings.port,
+          dbSettings.gphost, dbSettings.gppath, dbSettings.gpport, dbSettings.gpload,
+          IncrementalMode.MATERIALIZATION, null)
+
+        val schema = Map[String, VariableDataType]("r1.is_correct" -> MultinomialType(2))
```
netj: Shouldn't we test a multinomial type with more than two classes?

src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
feiranwang: The grounding code matches ddlog.jar within deepdive.

Previously, in the ddlog generated application.conf, we name weight column as "relation.relationAlias.columnName", where for incremental case relation needs to be prefixed with dd_delta_ or dd_new. This is very awkward.

Now, we use consistent naming for weight column. The name for weight column would be "dd_weight_column_idx", and is consistent for increnental or non-incremental run.

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
feiranwang: These lines were in master, so I used it here. I think incremental has been fixed for greenplum and postgresxl.

netj:
I wished the two fixes were in separate PRs, but if you think they're better together, I'm okay with either way.  Not 100% sure about all the details of multinomial incremental grounding part, but the code looks relatively okay.  We should really refactor and duplicate the code for different databases soon to stop introducing more branches to the control flow.

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
feiranwang: There's no difference in grounding for different multinomial domain sizes. Multinomial takes a different path from boolean in grounding factors and weights.

src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
netj: I see, I like that consistency through simplicity.  Actually, I was asking whether this change is harmless to the existing apps, not originating from ddlog.

src/main/scala/org/deepdive/datastore/MysqlDataStore.scala
feiranwang: Casting to signed in mysql should be also bigint. 
A related post is http://dba.stackexchange.com/questions/57363/why-does-cast1-as-signed-integer-return-a-bigint-on-mysql.

I'm installing mysql to confirm this.

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
netj: Hmm, then whoever fixed that forgot to update the tests.  Or it could be my fault while merging for releases.  Could you please remove these guards?

src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
feiranwang: No, this will not affect existing apps not originating from ddlog. It only affect incremental programs generated from ddlog.

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
netj: I know, but testing for two may not be a meaningful test case because it's the same as Boolean.

feiranwang:
Sorry, there are some refactoring in assigning weight ids, so I put the commits together. It's better if they were separate.

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
feiranwang: A multinomial with domain size two may feel like boolean, but it is not the same as boolean. The weights are indexed by variable cardinality. For example, if there's a `weight: ?`, for boolean, it's just one weight, but for multinomial, the number of weights is the variable domain size. 

This test has considered the semantics of multinomial, and the assert conditions are different from boolean case.

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
feiranwang: Sure

src/main/scala/org/deepdive/datastore/MysqlDataStore.scala
feiranwang: Locally, either casting 0 to unsigned or signed will become int typed...

src/test/scala/unit/inference/SQLInferenceDataStoreSpec.scala
netj: Right, what you said can be seen exactly from the test code.  However, the test seems too unrealistic (even the multinomial variable name is `is_correct`!) and makes the reader doubt whether this really is a valid test case.  I was just asking if it could be changed to look more real.  For example, increasing the number from 2 to 3, changing the name to class or category, and fixing the corresponding numbers in the assertions should make it completely look normal.  That's not a huge change, is it?

feiranwang: Sure, I'll do it. Sorry I just wanted to clarify the concepts... not worrying about the amount of changes...

feiranwang:
@netj Updated. Thanks!

netj:
LGTM

PR #348, opened 2015-08-05
Add support for greenplum to export to json

feiranwang:


netj:
I'll have to take a deeper look at the code soon, but could you add a test (.bats file) for this format=json part with the corner cases we manually checked?

feiranwang:
Is there an easy way to call `deepdive sql` inside .bats file?

netj:
That's a good question.  You can just run `deepdive sql` within a .bats file as long as it [sources the env.sh file](https://github.com/HazyResearch/deepdive/blob/master/test/test_plpy.bats#L4).  The [test/env.sh is setting up the PATH](https://github.com/HazyResearch/deepdive/blob/master/test/env.sh#L12) correctly for the staged `deepdive` command.

util/csvtojson.py
```diff
@@ -0,0 +1,59 @@
...
+def main():
+  # get column types
+  sql = "CREATE TEMP TABLE __to_json_temp AS (%s) LIMIT 0; \
+  COPY (SELECT format_type(atttypid, atttypmod) AS type \
+  FROM pg_attribute \
+  WHERE attrelid = '__to_json_temp'::regclass \
+  AND attnum > 0 \
+  AND NOT attisdropped \
+  ORDER BY attnum) TO STDOUT" %(sys.argv[1].rstrip(";"))
+  types = subprocess.check_output("deepdive sql %s" %pipes.quote(sql), shell=True).strip().split("\n")
```
netj: Nice! Looks like you found the magic SQL!

Calling `deepdive sql` within the implementation of itself seems pretty weird and inefficient.  I think this part that runs the fancy SQL should be moved out of python, and run directly from the `db-query`.  The types can be passed as arguments to this python script.  That way, we can also keep the python code much simpler, reusable, and having less dependencies.

util/csvtojson.py
```diff
@@ -0,0 +1,59 @@
...
+  reader = csv.reader(sys.stdin, delimiter = ',', quotechar = '"')
+  fieldnames = reader.next()
+  if len(types) != len(fieldnames):
+    raise ValueError("Number of columns does not match schema")
+  for line in reader:
+    if len(line) != len(fieldnames):
+      raise ValueError("Number of columns does not match schema")
+    obj = {}
+    for i in range(len(fieldnames)):
+      obj[fieldnames[i]] = convert_type(line[i], types[i])
```
netj: `xrange` is better if you're just iterating over it once as it won't create a temporary list. (It seems some consider [`range` as anti-pattern](http://lignos.org/py_antipatterns/).)  I'm unsure how fast `len` is, but keeping the number of columns in a local variable is probably a good idea.
This loop body is the hot path, so we should try our best to make it as efficient as possible.  Actually, instead of having `convert_type` go over multiple `if`-branches with `types[i]` on every input line, it should be significantly better to construct a list of functions based on the types ahead of time, to simply apply them on the values we see for each line without any branching.

shell/driver.greenplum/db-query
```diff
@@ -0,0 +1,28 @@
...
+case $header in
+    0) ;;
+    1) copy_option+=" HEADER" ;;
+    *) error "$header: unrecognized value for HEADER"
+esac
+case $format in
+    tsv) ;;
+    csv) copy_option+=" csv" ;;
+    json)
+        db-query "$sql" csv 1 | python $DEEPDIVE_HOME/util/csvtojson.py "$sql"
```
netj: You don't need to call the python script with full path as long as it stays in either util/ or bin/.  I don't think explicitly calling python is a good idea, either.  So, this can be written simply as:
```bash
db-query "$sql" csv 1 | csvtojson.py "$sql"
```

However, since we want to call the magic sql directly from this script and pass the types to python, it should be something like this:
```bash
types=$(db-prompt -c "CREATE TEMP TABLE ... ($sql LIMIT 0); ... magic SQL ...")
db-query "$sql" csv 1 | csvtojson.py $types  # XXX assuming no spaces in type names
```

util/csvtojson.py
```diff
@@ -0,0 +1,59 @@
+#! /usr/bin/env python
+
+import json, sys, csv, os, pipes, subprocess
```
netj: os, pipes, subprocess will be unnecessary once we take out the `deepdive sql` invocation.

netj:
Please see my comments.  I can help you rewrite the .py to construct a list of functions if you want.

feiranwang:
Thanks for the detailed comments! Looking into them.

feiranwang:
Updated.

test/greenplum/export_json.bats
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bats
+# Tests for json exporter
+
+. "$BATS_TEST_DIRNAME"/env.sh >&2
+
+@test "$DBVARIANT json exporter" {
+    q="SELECT 123::bigint as i, 45.678 as float, TRUE as t, FALSE as f, 'foo bar baz'::text as s, NULL::text as n, ARRAY['asdf qwer"$'\t'"zxcv"$'\n'"1234', '\"I''m your father,\" said Darth Vader.'] as a, ARRAY[1,2,3] as b"
+    expected="{\"a\": [\"asdf qwer\\tzxcv\\n1234\", \"\\\"I'm your father,\\\" said Darth Vader.\"], \"b\": [1, 2, 3], \"f\": false, \"i\": 123, \"float\": 45.678, \"n\": null, \"s\": \"foo bar baz\", \"t\": true}"
+    [[ $(db-query "$q" json 0) = "$expected" ]]
+}
```
netj: Let's keep this as a symlink to the postgresql one.

test/postgresql/export_json.bats
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bats
+# Tests for json exporter
+
+. "$BATS_TEST_DIRNAME"/env.sh >&2
+
+@test "$DBVARIANT json exporter" {
+    q="SELECT 123::bigint as i, 45.678 as float, TRUE as t, FALSE as f, 'foo bar baz'::text as s, NULL::text as n, ARRAY['asdf qwer"$'\t'"zxcv"$'\n'"1234', '\"I''m your father,\" said Darth Vader.'] as a, ARRAY[1,2,3] as b"
+    expected="{\"i\":123,\"float\":45.678,\"t\":true,\"f\":false,\"s\":\"foo bar baz\",\"n\":null,\"a\":[\"asdf qwer\\tzxcv\\n1234\",\"\\\"I'm your father,\\\" said Darth Vader.\"],\"b\":[1,2,3]}"
+    [[ $(db-query "$q" json 0) = "$expected" ]]
```
netj: Can we use `deepdive sql eval ... format=json` instead?

util/csvtojson.py
```diff
@@ -0,0 +1,64 @@
...
+    # string unescaping
+    # note this is an array as a csv field, we first unescape csv escaping,
+    # and then do array unescaping, apply csv escaping, and parse the arry
+    # elements as csv
+    # csv escapes double quote with two double quotes
+    # array escapes double quote using backslashes
+    flat_func = convert_flat_type_func(column_type)
+    def convert_text_array_func(value):
+      if value == "": return None
+      arr = csv.reader([value[1:-1].replace('""', '"').replace('\\"', '""')], delimiter=',', quotechar='"').next()
```
netj: I don't understand why you need the first csv unescaping?  Doesn't that mess up some cases?  How do you deal with escaped backslashes?  It'd be nice to include the corner case examples step by step in the comment.

feiranwang: When exporting to csv, array escape is first applied, and then csv escape is applied. To unescape, we need to do the reverse.

util/csvtojson.py
```diff
@@ -0,0 +1,64 @@
...
+      return [flat_func(x) for x in arr]
+    def convert_other_array_func(value):
+      if value == "": return None
+      arr = csv.reader([value[1:-1]], delimiter=',', quotechar='"').next()
+      return [flat_func(x) for x in arr]
+    if column_type == "text":
```
netj: I think you're shortcircuiting here for performance, right?  However, is it correct to do so for just text type?  Shouldn't the escape handling be the safe default and use the faster one only for the known types that are safe, e.g., int, float, double, number, etc.?

util/csvtojson.py
feiranwang: Will add unescaping for backslash.

test/postgresql/export_json.bats
```diff
@@ -0,0 +1,10 @@
...
+}
```
netj: Let's add some tests that also include backslashes, curly braces in the array.

util/csvtojson.py
feiranwang: Only text array needs unescaping, and int, float, boolean arrays do not need unescaping. Here I'm testing if it's a text array

shell/driver.greenplum/db-query
```diff
@@ -0,0 +1,35 @@
...
+    csv) copy_option+=" csv" ;;
+    json)
+        types="$(db-prompt -c "CREATE TEMP TABLE __to_json_temp AS ($sql) LIMIT 0;
+            COPY (SELECT format_type(atttypid, atttypmod) AS type
+            FROM pg_attribute
+            WHERE attrelid = '__to_json_temp'::regclass
+            AND attnum > 0
+            AND NOT attisdropped
+            ORDER BY attnum) TO STDOUT" | tr '\n' ',' | sed 's/,$//')"
+        db-query "$sql" csv 1 | csvtojson.py "$types"
```
netj: No need to turn the whole thing into a single comma-separated string.  I think it's better to pass the types as multiple arguments without quoting `$types` here.

test/postgresql/export_json.bats
feiranwang: `deepdive sql` command requires `db.url`, `schema.sql`, etc., which are not present for tests.

util/csvtojson.py
```diff
@@ -0,0 +1,64 @@
...
+      return convert_text_array_func
+    else:
+      return convert_other_array_func
+  else:
+    return convert_flat_type_func(column_type)
+
+def main():
+  # get column types
+  types = sys.argv[1].split(",")
```
netj: Let's use multiple arguments.
```python
types = sys.argv[1:]
```

test/greenplum/export_json.bats
feiranwang: postgresql and greenplum's output is different, postgresql's output does not have spaces after the key colon, and value delimiter.

shell/driver.greenplum/db-query
feiranwang: Sure

util/csvtojson.py
netj: But isn't the `csv.reader` supposed to handle the first csv escapes?

test/greenplum/export_json.bats
netj: Hmm.. that sounds like a weird reason to keep the two separate.  Since the whitespaces outside string values are not the test's concern, it should rather normalize the json before checking, right?

util/csvtojson.py
feiranwang: Here, the array is inside the csv file, so the array itself is double quoted. Every double quote inside the array is doubled. After we unescape the csv and psql array, the elements of array become normal csv escaped elements.

test/postgresql/export_json.bats
netj: I see.  Sounds reasonable for now.  I should relax that requirement at some point.

test/greenplum/export_json.bats
feiranwang: I searched a bit, and it seems there is no easy way to normalize json in bash...

netj: Yeah, python is probably a good choice for that and bash is good friend with python.  Actually I just updated the tests myself.  Hope it's helpful.

test/postgresql/export_json.bats
netj: Done.

feiranwang:
Thanks for helping on this! I updating the rest.

util/csvtojson.py
netj: Please take a look at https://github.com/HazyResearch/mindbender/blob/3f0e24ed0851907302c49ebe1c1d619e7c243df2/gui/frontend/src/mindtagger/array-parsers.coffee#L10
You'll see some incorrect cases with the updated tests.

netj:
It seems earlier versions of postgres also needs this workaround.  Travis tests are broken which has psql 9.1: https://travis-ci.org/HazyResearch/deepdive/builds/74359333#L2436-L2439
I think we may check the postgres version to see if `to_json()` is supported and decide which approach to use, instead of keeping this as a greenplum specific extension.  The version check should ideally be done once from `db-parse` to set some env vars that controls `db-query`.

feiranwang:
Instead of writing postgresql version-specific code, can we just unify this by using our own script?

netj:
I'll take care of the version check once you finish greenplum's `db-query`.

feiranwang:
Thanks! Pushed my commit.

netj:
I'm looking into distinguishing empty string vs null with csv.reader.  Please let me know if you have a good idea.  Also any comment on my new code is welcome!

feiranwang:
Good point. I think that's hard, as the csv is the same...

netj:
No, psql produces a clearly different csv: empty strings are quoted (...,"",...), but NULLs aren't: (...,,...), as I commented in the csvtojson.py.

feiranwang:
I see...

netj:
My tests probably got way too extreme, but I think it's worth spending some time once and for all getting this part perfectly reliable. Unfortunately our script cannot handle all the corners, but the limitations are at least clearly written in the tests, and hopefully they'll affect very few users who use old versions. 

Please take a look and let's get this in unless you have concerns.

feiranwang:
Looks good to me. The tests are comprehensive. 

Not related to this PR, mysql spouse tests fails from time to time. I didn't find obvious problems yet.

netj:
Thanks for taking a look.  You're right.  Tests fail on mysql spouse example quite often.  Not sure why though.  Can you put up an issue or look into it?

feiranwang:
Sure.

Issue #329, opened 2015-06-23
Refactor Scala code


netj:
1. Remove clutter from Scala
    * [ ] Rewrite calibration plot generator in a separate post-processing script and remove org.deepdive.calibration._
    * [ ] Remove org.deepdive.settings.models._ and have the rest of the code use Config.get* directly as much as possible
    * [x] Rewrite SettingsParser.scala to simply validate and fallback to default values for
        * [x] db.default, schema, calibration, sampler, pipeline
        * [x] every extraction.extractors
        * [x] every inference.factors
        * [ ] except for providing some utilities
    * [ ] Drop json_extractor support

2. Remove Akka by emitting an execution "Plan" instead of immediately executing SQL queries and commands
    * [ ] Add a Plan class to hold SQL queries and arbitrary commands
    * [ ] Add easy ways to emit SQL queries and commands
    * [ ] Add an easy way to group them to create an hierarchy
    * [ ] Add an easy way to declare dependency among them
    * [ ] Rewrite org.deepdive.inference._ and .extraction._
    * [ ] Write a code generator that translates a Plan holding a DAG of SQL/commands into a shell script
    * [ ] Have `deepdive` command first generate the shell script, then execute it

3. Decouple database drivers
    * [ ] Duplicate InferenceRunner implementation for every backend, inlining *Datastore: postgresql, greenplum, postgresql-xl, mysql, mysqlcluster
    * [ ] Duplicate ExtractorRunner for every backend as well

chrismre:
+1

PR #353, opened 2015-08-12
Refactored Scala code for settings

netj:
- Uses named arguments syntax in constructors.
- Uses `.copy()` with named arguments when possible.
- Splits unit tests into individual .bats files, so each can be run/excluded selectively.

(Part of #329)


chrismre:
Is this the end of akka?

netj:
Not yet, but paving the way for it!

src/main/scala/org/deepdive/settings/SettingsParser.scala
```diff
@@ -1,253 +1,536 @@
...
+
+    log.info(settings.config.root.render(ConfigRenderOptions.concise().setFormatted(true)))
+    log.info(settings toString)
+
+    settings
   }
 
+  private def loadDatabaseSettings(settings: Settings): Settings = {
+    var config = settings.config withFallback ConfigFactory.parseString(
+      """
```
feiranwang: What is this string used for?

feiranwang: Oh I got it. `withFallback` combines this string with the input config.

src/main/scala/org/deepdive/settings/SettingsParser.scala
```diff
@@ -1,253 +1,536 @@
 package org.deepdive.settings
```
feiranwang: Is it better to combine the settings in one file or in separate files? It seems after combining, this file is getting big and comprehensive.

src/test/scala/helpers/TestHelper.scala
```diff
@@ -35,18 +35,20 @@ object TestHelper {
...
-          System.getenv("DBPASSWORD"),    // password
-          System.getenv("DBNAME"),        // dbname
-          System.getenv("DBHOST"),
-          System.getenv("DBPORT"),
-          System.getenv("GPHOST"),
-          System.getenv("GPPATH"),
-          System.getenv("GPPORT"),
-          getGPLOADEnv(),
-          getIsIncrementalEnv())
+      DbSettings(
```
feiranwang: (Not related to this PR)
I think we might want to use a map for `DbSettings`. When adding new settings inside `DbSettings`, we need to to change all the initializations for the case class. A map may be much easier to maintain.

src/test/scala/helpers/TestHelper.scala
```diff
@@ -35,18 +35,20 @@ object TestHelper {
...
+        user = System.getenv("DBUSER"), // user
+        password = System.getenv("DBPASSWORD"), // password
+        dbname = System.getenv("DBNAME"), // dbname
+        host = System.getenv("DBHOST"),
+        port = System.getenv("DBPORT"),
+        gphost = System.getenv("GPHOST"),
+        gppath = System.getenv("GPPATH"),
+        gpport = System.getenv("GPPORT"),
+        gpload = getGPLOADEnv(),
+        incrementalMode = getIsIncrementalEnv()
```
feiranwang: Great! Named arguments are much clearer.

src/main/scala/org/deepdive/settings/SettingsParser.scala
netj: Larger functions are no doubt bad, but I don't think that necessarily applies to files.  I found it much easier to stare at one file than getting lost while trying to guess where something might exist.  Since Scala doesn't force us to sync the filenames with its contents, I think it's better to keep the number of files minimal.  The only downside I see is slower compilation due to putting too many independent classes in the same file.  But I believe settings classes don't need frequent changes.  Does this sound convincing?

src/test/scala/helpers/TestHelper.scala
netj: I'm against using Map.  I initially thought we should directly use the Config object instead of creating an array of Settings classes.  But it seems Config is awkward to use (throws exception for null) and since we already have this typed intermediate representation it's good to maintain it cleanly.
Using case classes with named argument syntax and default values eliminates the need for impacting all constructors.  And right, using `.copy()` on top of that further reduces the impact.  This PR is actually all about these two to make it easier to further refactor things.

src/test/scala/helpers/TestHelper.scala
netj: Sure, everyone should use this style unless there are only one or two args.

feiranwang:
Looks good. Merging.

PR #364, opened 2015-08-21
Makes deepdive shell robust from user's old `$APP_HOME`

netj:
by using a new name `$DEEPDIVE_APP` and only setting the legacy environment variable, not reading it.

Some users set such variables in their shell or source the env.sh which can cause trouble.


feiranwang:
Looks good. Merging.

PR #365, opened 2015-08-22
Improves installers

netj:
* Adds Debian support
* Shows more friendly and verbose messages when run on unsupported OSes
* Lets user keep going on unrecognized versions of OS
* Adds capability to download source tree and run tests against installed versions
* Fixes some bugs in shell, db drivers

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -38,10 +38,35 @@ init_INSTALLER_TEMP_DIR() {
 
 # common installers ###########################################################
 # installs DeepDive's build dependencies
-install__deepdive_build_deps() { false; }
-install__deepdive_build_deps() { false; }
+install__deepdive_build_deps() {
+    local deps=
+    deps=(
+        git
+        make
+        bzip2
+        unzip
+        jdk
+    )
+    error "Please make sure the following packages are available on your system: $(
+        printf '\n  %s' "${deps[@]}")" ||
+            [[ ! -t 1 ]] || read -p "# Continue? " || true
+}
 # installs DeepDive's runtime dependencies
-install__deepdive_runtime_deps() { false; }
+install__deepdive_runtime_deps() {
+    local deps=
+    deps=(
+        bash
+        coreutils
+        xargs
+        bc
+        gnuplot
+        jre
+        python
+    )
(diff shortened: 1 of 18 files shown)


util/install.sh
```diff
@@ -102,6 +148,15 @@ case $(uname) in
         #elif [[ -e /etc/redhat-release ]]; then
         #    # CentOS/RedHat
         #    os=RedHat
+        else # unrecognized Linux distro
+            os=
+            # try to grab a Linux distro identifier
+            {
+                set -o pipefail
+                hint=$(lsb_release -i | cut -f2) ||
```
feiranwang: Will there be a problem if `lsb_release` or `/etc/redhat-release` does not exist?

util/install/install.Ubuntu.sh
```diff
@@ -1,53 +1,40 @@
 #!/usr/bin/env bash
-# DeepDive installers for Ubuntu Linux
+# DeepDive installers for Debian/Ubuntu Linux
 
-LSB=$(lsb_release -r 2>/dev/null)
-case ${LSB##*	} in
-    12.04|14.04|15.04) true ;;
-    *) error "Ubuntu $LSB found: This installer only works with Ubuntu 12.04, 14.04, and 15.04."
+LSB=$(lsb_release -ir | cut -f2)
```
feiranwang: On raiders4 and raiders5 (Ubuntu 12.04), I got `No LSB modules are available` when trying `lsb_release`. Should we deal with `lsb_release` before using it?

netj: I think you need to add `-ri` flags to get normal output.  `lsb_release` seems to be always available on Debian-based systems.  Most base packages are depending on it, so should be installed.  Btw when `lsb_release` goes wrong, the current code will just show warnings and still let the user run the defined installers.  Maybe I should make it more explicit by adding a ` || true`.

util/install.sh
netj: Good point.  The intention was to then set hint empty and just display a warning without it.  It seems the second line may fail if `redhat-release` doesn't exist.  I'll bulletproof it.

feiranwang:
Looks good. Merging.

PR #368, opened 2015-08-25
Add deepdive-load for tsv and csv

feiranwang:
`deepdive load` command, supports tsv and csv.

### DIFF
--- a/shell/deepdive-load
+++ b/shell/deepdive-load
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# deepdive-load -- Loads a table from a file or Pipe's output.
+# You must either have the DEEPDIVE_DB_URL environment set to a proper URL or
+# run this command under a DeepDive application where the URL is set in a db.url file.
+#
+# > deepdive load table file format
+#
+# > deepdive load sentences sentences.tsv tsv
+# > deepdive load sentences sentences.csv csv
+#
+##
+set -eu
+
+# parse database settings
+. load-db-driver.sh
+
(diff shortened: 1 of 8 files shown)


shell/deepdive-load
```diff
@@ -0,0 +1,28 @@
...
+    table=$1; shift
+    file=$1; shift
+    if [[ $# -gt 0 ]]; then
+        format=$1; shift
+    else
+        format=tsv
+    fi
+    db-load $table $file $format
+else
+    error "Missing table name or file name"
```
netj: Please move this usage error checking to the top to avoid unnecessary global if blocks:
```bash
# ...
. load-db-driver.sh

[[ $# -gt 1 ]] || error "Missing table name or file name"

table=$1; shift
# ...
```

shell/driver.postgresql/db-load
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+# db-load -- Loads file to the PostgreSQL database configured for a DeepDive application
```
netj: I think you need to add this new `load` operation to `load-db-driver.sh` as well.

netj:
Please see my minor comments.  I want `deepdive load` to support more formats (e.g., Bazaar/Pipe's json columns) and implicit data locations for tables under `input/`, but this is a good start.

feiranwang:

feiranwang:
Updated.

netj:
Looks good.  Seems like the python script is not necessary.

PR #370, opened 2015-08-27
Smart split for extractor input based on parallelism

zifeishan:
It has been a problem for users to set input_batch_size to split extractor input correctly. Too small chunk size will yield too many udf executions and many file copy operations, which can be slow. Too large chunk size will reduce parallelism to 1.

Splitting the extractor input by parallelism should be the way to go. In this version, it detects whether system supports "split -n" command which split files into given number of chunks. If not supported, it falls back to use input_batch_size.

### DIFF
--- a/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
+++ b/src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
@@ -460,7 +460,7 @@ class ExtractorRunner(dataStore: JdbcDataStore, dbSettings: DbSettings) extends 
       Helpers.executeCmd(splitCmdByPar);
     } catch {
       case e: Throwable =>
-        log.debug(s"Split by number of chunks is not supported. Try to use batch size ${linesPerSplit} to split...") 
+        log.debug(s"Split by number of chunks is not supported. Try to use batch size ${linesPerSplit} to split...")
         executeScriptOrFail(splitCmd, taskSender)
     }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
```diff
@@ -449,12 +449,20 @@ class ExtractorRunner(dataStore: JdbcDataStore, dbSettings: DbSettings) extends
     log.info(s"File dumped to ${actualDumpedPath}")
     val splitPrefix = s"${actualDumpedPath}-"
     val linesPerSplit = task.extractor.inputBatchSize
+    val maxParallel = task.extractor.parallelism
+    val splitNum = maxParallel * 10
```
netj: Why do we need `* 10` here?  If I understood correctly, your fix is trying to avoid the startup overhead of UDFs, right?  Then why do we still pay 9 more times if we can just start once and handle the chunk?

netj:
Looks like a nice workaround for an inefficiency in our current extractor runner.  Please see my inline comment.

src/main/scala/org/deepdive/extraction/ExtractorRunner.scala
zifeishan: @netj Sorry for the late response. Here I was trying to reduce the effect of the data skew:  sometimes one thread might be stuck in a chunk of data and all other threads are waiting for the slowest one to finish. With smaller splits the data skew should have less impact.  So the idea here is to find a balance between UDF start overhead and the data skew.

I also believed that `xargs` has some load balancing to distribute the input chunks but I am not sure there.

zifeishan:
@netj Should we merge this? I've addressed the comments by setting the number of splits to parallelism. Thanks!

netj:
@zifeishan I'm a bit reluctant since we'll address this in a new codebase, but I'm fine merging if you guys need this fix in the short term.

PR #376, opened 2015-09-04
Ddlog update

feiranwang:
1. Fix delta deriver for incremental function call, revise syntax.
2. Add support for globally auto-set parallelism for extractors.
3. `deepdive initdb TABLE` command for initializing a single table.


chrismre:
@feiranwang Is this waiting @netj approval? He should be assigned :)

shell/deepdive-run
```diff
@@ -91,6 +91,15 @@ fullConfig=$run_dir/deepdive.conf
     ddlog compile "${ddlogFiles[@]}"
     export PIPELINE=  # XXX ddlog shouldn't emit this
     : ${Pipeline:=endtoend}
+
+    # set PARALLELISM env var, use max parallelism if the variable is not set
+    if [[ $(uname) = 'Linux' ]]; then
```
netj: Checking if `nproc` or `sysctl` is available makes more sense than relying on uname.  You could chain the options with something like:
```bash
: ${PARALLELISM:=$({
    # Linux typically has coreutils which includes nproc
    nproc ||
    # OS X
    sysctl -n hw.ncpu ||
    # fall back to 1
    echo 1
} 2>/dev/null)}
```

netj:
Doesn't `deepdive initdb TABLE` still drop and create the whole database before creating the given table, affecting others?  Here's what I think users expect from the initdb command:

* When there are arguments, DD should drop/create/load just those specified tables.  Assuming it's a DDlog app, DD should drop/create the tables from the DDlog schema, then optionally load data to the new tables from an assumed path under input/ by some naming convention.  It's an error if it's not a DDlog app.
* When no argument is given, DD should drop/create all known tables.  If it's a DDlog app, all tables defined in the schema should be created then loaded as if the names were all given manually.  If it's not DDlog, it should rely on schema.sql and input/init.sh to initialize the database.  For this last non-DDlog case, DD should perhaps do a dropdb to be backward compatible.

In any case, DD should first make sure the database is created.

shell/deepdive-initdb
```diff
@@ -16,8 +16,19 @@ db-init "$@"
 
 # make sure the necessary tables are all created
 if [[ -e app.ddlog ]]; then
-    # TODO export schema.sql from ddlog instead of running initdb pipeline
-    deepdive-run initdb
+    if [[ $# -gt 0 ]]; then
```
netj: Rather than having these argument count checks buried deep inside, I think it's much clearer to define initdb's behavior entirely differently when arguments are specified.  Please see my comment on the PR for reorganizing.

util/ddlog_initdb
```diff
@@ -0,0 +1,23 @@
+#! /usr/bin/env python
+# Generate create table statement given a ddlog exported schema and a table name.
+# Usage: ddlog_initdb SCHEMA.JSON TABLE_NAME
```
netj: Let's give a better name to this script.  How about `schema_json_to_sql`?

I was originally thinking this SQL generator should go under each driver, e.g.,  to handle `DISTRIBUTED BY` in GP and so on.  If you agree, I think it'll be a matter of just moving this to `shell/driver.postgresql/` and keeping a symlink or clone under `driver.mysql/`.

test/postgresql/deepdive_initdb.bats
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bats
+# Tests for initdb
```
netj: Maybe it's more important to test the python script for generating CREATE TABLE statements?

netj:
Nice updates.  Please see my comments.

feiranwang:

feiranwang:
@netj Updated. Thanks!

netj:
Looks good, merging.

PR #380, opened 2015-09-07
SSL support

SenWu:
Added SSL support.
In order to support ssl connections. You can specify this in the connection string with:

    jdbc:postgresql://localhost:5432/dbname?ssl=true
Then Deepdvie will run SSL connection with the non-validating ssl factory:

    jdbc:postgresql://localhost:5432/dbname?ssl=true&sslfactory=org.postgresql.ssl.NonValidatingFactory

### DIFF
--- a/shell/driver.postgresql/db-parse
+++ b/shell/driver.postgresql/db-parse
@@ -24,7 +24,7 @@ echo "DEEPDIVE_JDBC_DRIVER=org.postgresql.Driver"
 if [ ${ssl} = "true" ]; then
     echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname?ssl=true\&sslfactory=org.postgresql.ssl.NonValidatingFactory"
 else
-    echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname" 
+    echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname"
 fi
 echo "export DEEPDIVE_JDBC_DRIVER DEEPDIVE_JDBC_URL"
 
--- a/shell/driver.postgresql/db-parse
+++ b/shell/driver.postgresql/db-parse
@@ -21,7 +21,11 @@ echo "export DBHOST DBPORT DBUSER DBPASSWORD DBNAME"
 
 # more variables for Scala application.conf
 echo "DEEPDIVE_JDBC_DRIVER=org.postgresql.Driver"
-echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname"
+if [ ${ssl} = "true" ]; then
+    echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname?ssl=true\&sslfactory=org.postgresql.ssl.NonValidatingFactory"
+else
+    echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname" 
+fi
 echo "export DEEPDIVE_JDBC_DRIVER DEEPDIVE_JDBC_URL"
 
 # XXX several environment variables for legacy apps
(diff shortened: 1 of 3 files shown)


shell/parse-url.sh
```diff
@@ -18,4 +18,7 @@ port=${port#:}
 user=${user_password%:*}
 password=${user_password#$user}
 password=${password#:}
-dbname=${rest}  # TODO do we need to strip any parameters from the rest of the URL?
+dbname=${rest%\?*}
+rest1=${rest#$dbname?}
+ssl=${rest1#ssl=} # TODO we only support ssl parameter from the rest of the URL
```
netj: This is fine for now, but spending a bit more effort to handle general query string parsing may be a wise move.  Well, even without doing any smart parsing, I think it may be more robust to pass the entire query string down to the JDBC url (say, via `$querystring`) and set the `PGSSLMODE=require` from db-parse if it contains `*ssl=true*`.  How does that sound?

shell/driver.postgresql/db-parse
```diff
@@ -21,7 +21,11 @@ echo "export DBHOST DBPORT DBUSER DBPASSWORD DBNAME"
 
 # more variables for Scala application.conf
 echo "DEEPDIVE_JDBC_DRIVER=org.postgresql.Driver"
-echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname"
+if [ ${ssl} = "true" ]; then
+    echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname?ssl=true\&sslfactory=org.postgresql.ssl.NonValidatingFactory"
+else
+    echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname"
+fi
```
netj: If the entire query string were available as `$querystring`, no need for duplicating the long full URLs:
```bash
echo "DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname$querystring"
```

I think the `&sslfactory=org.postgresql.ssl.NonValidatingFactory` option should be moved to documentation.  This is for getting self-signed certs to work, right?

shell/driver.postgresql/db-parse
```diff
@@ -29,4 +33,9 @@ echo "PGHOST=${host}"
 echo "PGPORT=${port}"
 echo "PGUSER=${user}"
 echo "PGPASSWORD=\$DBPASSWORD"
-echo "export PGHOST PGPORT PGUSER PGPASSWORD"
+if [ ${ssl} = "true"  ]; then
+    echo "PGSSLMODE=require"
+else
+    echo "PGSSLMODE=disable"
+fi
```
netj: Using the `$querystring`, we could do something like:
```bash
case $querystring in
    *ssl=true*) echo "PGSSLMODE=require" ;;
    *) echo "PGSSLMODE=disable"
esac
```

shell/parse-url.sh
```diff
@@ -18,4 +18,7 @@ port=${port#:}
...
+#dbname=${rest}  # TODO do we need to strip any parameters from the rest of the URL?
```
netj: Please delete this last line instead of commenting out.

netj:
Thanks for your quick fix.  Please see my comments to make it more robust to future changes.

SenWu:

shell/driver.postgresql/db-parse
SenWu: Sorry, I don't understand 'I think the &sslfactory=org.postgresql.ssl.NonValidatingFactory option should be moved to documentation.'. 
Yes,  `&sslfactory=org.postgresql.ssl.NonValidatingFactory` is for getting self-signed cert, and we add it in to `DEEPDIVE_JDBC_URL` for that purpose.

SenWu:
Updated.

shell/driver.postgresql/db-parse
netj: I was saying user should decide whether to disable certificate validation, not DeepDive. If we always disable validation, it's not real ssl support.

SenWu: OK. Should I just keep `querystring` in `DEEPDIVE_JDBC_URL` , e.g., ``echo"DEEPDIVE_JDBC_URL=jdbc:postgresql://$host${port:+:$port}/$dbname$querystring"`?

netj:
It seems the Postgres tests are all skipped perhaps because of some parsing issue?

shell/driver.postgresql/db-parse
netj: Yup. I think that's better because user can supply extra options we are not aware of when necessary.

SenWu: Sorry about that. Fixed.

SenWu:
Update docs. Please check, thanks!.

Issue #378, opened 2015-09-06
PGXL failing on large datasets


raphaelhoffmann:
Trying to run one of our memex extractors, PGXL fails on the following query:

SELECT DISTINCT t0.id AS factor_id, t1.id AS weight_id,  "ismassageparlorad.id"
  FROM dd_query_f_ismassageparlorad_features t0, dd_weights_f_ismassageparlorad_features t1
  WHERE  t0."ismassageparlorad.feature" = t1."ismassageparlorad.feature"

A few observations:
- It seems DISTINCT is redundant and can be dropped (?)
- I added an index on the two feature columns, but that didn't help
- feature is not a distribution column – we can try to redistribute the table (but shouldn't that happen automatically, when trying to join?)

alldefector:
PGXL distributes by first usable (primitive types) column by default. If both base tables were distributed by the ID column, this join would probably make each data node send data to all other data nodes -- by the same token, each node would also process data from all other nodes. And the cross-node data pipes might not be very efficiently implemented.

Ideally the DDL code generated by DD should specify the best dist key (by feature in this case).

raphaelhoffmann:
dd_query_f_ismassageparlorad_features (38GB) is distributed by "ismassageparlorad.id".
dd_weights_f_ismassageparlorad_features (18MB!) is distributed by REPLICATION.

This join should be trivial, no matter how the first table is distributed.

We also can't re-distribute the table for every join, because re-distribution is very expensive. I'm currently re-distributing the first table by feature, but the process is CPU-bound with only the coordinator showing activity (100% CPU).

Maybe I'm using too many nodes (32).

raphaelhoffmann:
I tried re-running the query after redistributing the first table.

It's running now on our Azure instance using PGXL at a rate of 3.8MB/sec (this doesn't include the time for redistribution which itself took maybe an hour or more).

I also checked my logs to see how a corresponding query executed on Greenplum on raiders4. There, the rate is 1.8MB/sec (but we don't do redistribution). 

Not sure why this takes so long using either database system. The data is so small that it could easily fit in memory.  

alldefector – when you get a chance can you log into the machine and see if there's an easy fix? Thanks.

raphaelhoffmann:
Here are the query plans:

Greenplum (after ANALYZE):
```
 Gather Motion 48:1  (slice2; segments: 48)  (cost=106201580.45..115136401.10 rows=11701014 width=24)
[...]

PR #382, opened 2015-09-07
fix DISTINCT issue in grounding

SenWu:
Fixed DISTINCT issue mentioned #378

### DIFF
--- a/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
+++ b/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
@@ -715,8 +715,10 @@ trait SQLInferenceRunner extends InferenceRunner with Logging {
         execute(dataStore.analyzeTable(querytable))
         execute(dataStore.analyzeTable(weighttableForThisFactor))
 
+        // do not need DISTINCT in this sql, since for every tuple in the querytable,
+        // there is only one matching tuple in the weighttableForThisFactor
         du.unload(s"${outfile}", s"${groundingPath}/${outfile}", dbSettings,
-          s"""SELECT DISTINCT t0.id AS factor_id, t1.id AS weight_id, ${idcols}
+          s"""SELECT t0.id AS factor_id, t1.id AS weight_id, ${idcols}
            FROM ${querytable} t0, ${weighttableForThisFactor} t1
            ${weightJoinCondition};""", groundingDir)
 


netj:
Any anticipated side effects if we drop DISTINCT?  I think there must have been a good reason or use case why we added it in the first place.

chrismre:
This makes me really nervous :)

alldefector:
I wasn't able to find who first introduced DISTINCT with git blame because
of some refactoring. Maybe the initial author just has to speak up about
his thought process... Or someone with better git skills could track it
down...

chrismre:
If DISTINCT is not needed, then there has to be some convention somewhere
that ensure DISTINCT is unnecessary. IF you want to remove it, explain what
that convention and make sure it's documented (and ideally checked!)

raphaelhoffmann:
Yes, some more documentation would be helpful here. Here's why I think that DISTINCT is not necessary. Our query is 
https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L719

*Claim 1: For every tuple in the left table, there is at most one join tuple.*

The join condition is a key for the right table. 

https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L710
https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L688
https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L579

*Claim 2: Each join tuple is distinct.*

The join selects factor_id which is a key for the left table.

https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L563

zhangce:
@raphaelhoffmann You are right, for the following query

SELECT DISTINCT t0.id AS factor_id, t1.id AS weight_id, "ismassageparlorad.id"
FROM dd_query_f_ismassageparlorad_features t0, dd_weights_f_ismassageparlorad_features t1
WHERE t0."ismassageparlorad.feature" = t1."ismassageparlorad.feature"

The second relation (`dd_weights_f_ismassageparlorad_features`) is grounded by running DISTINCT
over the first relation (`dd_query_f_ismassageparlorad_features`). Therefore, there is no way this query can generate duplicate result.

@SenWu Can you document this in the code?

Ce

SenWu:
@zhangce Yes, I found I add it in our incremental update which it is unnecessary. Sorry about that. Will add specification in the code.

SenWu:
Updated.

chrismre:
@SenWu @zhangce We have some issues failing on the pr... can we figure out the issue here? How robust is the failing incremental test? 

I'm, for example, worried that it would fail on another architecture or with different random draws, etc. Can someone give me some better understanding of this test? (in meeting is fine)

feiranwang:
@chrismre There was a bug introduced by an earlier commit, and it is fixed in #376. The incremental test is stable after merging #376.

chrismre:
sounds good. Thanks for the explanation!

PR #390, opened 2015-09-17
Fix DDLog source snippets

AlexisBRENON:
Every snippets were truncated (missing first line). Update script calls to properly display each code block.


AlexisBRENON:
Maybe there is other pages with code snippets that need to be fixed. Let me know, I will do it.

netj:
Thanks for the fix!  Ideally, these line numbers should be auto generated based on some markers at some point.

Issue #336, opened 2015-07-14
Update documentation TOC


netj:
We need to reorganize the table of contents for the documentation.
* Many items in Basics seem too Advanced.
* Tutorial should come right after Background.
* Many shorter pages should be merged into larger ones or simply rewritten into a section in FAQ.
* Each group can have more structure: one for various database backends, another for references, etc.
* We can get rid of basics/ and advanced/ under doc/doc/.  Perhaps there's no meaningful reason to bake such arbitrary grouping (Background, Basics, Advanced, Tutorial) into the source layout.

PR #395, opened 2015-09-18
Reorganized documentation TOC

netj:
with nesting to make the structure more obvious.

Didn't rename any files yet as planned in #336, which can be done later before the next release.


SenWu:
LGTM.

PR #397, opened 2015-09-20
Ce fix plpython dependency

zhangce:
@netj If you could review this that would be great. Thanks!


src/main/scala/org/deepdive/datastore/PostgresDataStore.scala
```diff
@@ -198,7 +203,7 @@ class PostgresDataStore extends JdbcDataStore with Logging {
           WHERE (host, port) IN (SELECT host, min(port) FROM pgxl_dual group by host);
       """)
     }
-    executeSqlQueries(SQLFunctions.piggyExtractorDriverDeclaration, false)
+    //executeSqlQueries(SQLFunctions.piggyExtractorDriverDeclaration, false)
```
netj: Can you delete this line instead of commenting?

netj:
Thanks for the fix!  Looks good other than that minor one.

zhangce:
@netj Thanks, I integrated your feedback.

Issue #407, opened 2015-09-24
Open data in a better format than .zip?


netj:
.zip files need to be fully downloaded for any of its files to be extracted, which renders any partial download useless.
If the datasets were published in .tar.gz with smaller files (e.g., README) at the beginning, then it's possible to extract whatever files that exist in the first part of the archive.

zhangce:
@netj is this really necessary? The goal of the teaser data set is to make sure the user can download a small sample instead of the whole file... Can you let me know more details why the teaser data set is not enough such that the user need to download partial full data set with this trick?

If there is a need to download partial dataset larger than the teaser but smaller than the full, I can add a larger teaser in the middle.

netj:
I partly agree this isn't crucial with the teaser.  In my case, I was trying to confirm the whole dataset's format was identical to the teaser, but it was too large to download.  I thought we could switch to a superior format because we're going to republish them anyway for fixing other bugs.

zhangce:
>> because we're going to republish them anyway for fixing other bugs.

Only 2 out of 8 datasets need republishing. I will make the new ones
.tar.gz. (I personally like zip better but that's not important :)) For the
others, I will leave them untouched for now if it does not impact user
experience.

Ce

Issue #406, opened 2015-09-24
Open data READMEs are out of sync and need more description


netj:
* [ ] The TSV formats contain lemma columns but they do not appear in the README.
* [ ] Some people were asking for a one-sentence description of each column.
* [ ] Adding hints for creating indexes over which columns for speeding up DeepDive to the README would be useful.

zhangce:
@netj Maybe these three should happen in the walkthrough? When I 
describe the data format in the open data page, I pointed it to the 
walkthrough...

thodrek:
@netj @zhangce I'll do a pass over the documentation as well. Starting now.

zhangce:
@thodrek Sorry for the confusion, just to be clear, the open data doc is not the core part--you can focus on the main doc (I can deal with everything related to open data `opendata/index.md`)

netj:
@zhangce Yes, perhaps adding in the README a link to the right section of the tutorial or even to the opendata page is enough.  Once people download these datasets, they seem to get detached from the web page and expect things to be self-contained.

zhangce:
I see, I will make it clear in the README where to look up. Thanks!

Ce

Issue #405, opened 2015-09-24
Open data provenance columns are malformed


netj:
The provenance columns in TSV of most [open datasets](http://deepdive.stanford.edu/doc/opendata/) seem to be incorrect and some of them are malformed causing syntax errors when loading into the database.  The CoNLL format seems to contain the correct START/END offsets, so this seems to be a bug in conversion to TSV.
```
$ deepdive sql "COPY sentences FROM STDIN" <./input/bmc_full_dddb/sentences-0.tsv
ERROR:  malformed array literal: "{,/,2/,2/,1/,6/,7/1,24/1,26/1,28/1,43/1,51/1,60/1,64/1,69/1,76/2,07/2,86/2}"
DETAIL:  Unexpected "," character.
CONTEXT:  COPY sentences, line 1, column bounding_boxes: "{,/,2/,2/,1/,6/,7/1,24/1,26/1,28/1,43/1,51/1,60/1,64/1,69/1,76/2,07/2,86/2}"
```

zhangce:
@netj The fixed run started, ETA tomorrow afternoon for all datasets. Sorry for the bug here.

PR #414, opened 2015-09-26
fix bug/doc in open data

zhangce:
@netj These are the version that fixed issues
https://github.com/HazyResearch/deepdive/issues/407
https://github.com/HazyResearch/deepdive/issues/406
https://github.com/HazyResearch/deepdive/issues/405

Sorry for the bug!

Ce

### DIFF
--- a/doc/doc/opendata/index.md
+++ b/doc/doc/opendata/index.md
@@ -116,7 +116,7 @@ Each dataset is versioned with date, and with the MD5 checksum in the file name.
           <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-expanded="false"> Download Full Corpus <span class="caret"></span>
           </button>
           <ul class="dropdown-menu" role="menu">
-            <!--<li><a href="http://i.stanford.edu/hazy/opendata/pmc/pmc_teaser_dddb_20150131_3eb5601f6acecd58527a0dc6e55d0de7.zip">DeepDive-ready DB Dump</a></li>-->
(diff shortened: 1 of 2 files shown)


netj:
Thanks for making these fixes.

I think the dataset README is still confusing to just point to the general walkthrough.  How about pointing to the opendata's "Data Format (NLP Markups)" section and enumerating the list of columns there under the first bullet?  In fact, the opendata TSVs are slightly different from the walkthrough: they have provenance and word index columns, sentence id appears after document id, the dependency paths are represented differently using two columns.  I think these datasets are often consumed/prepared by people who have no idea/interest in what they're processing or how they are actually used in DD, so I think we need to be very precise in describing the exact schema.

zhangce:
Thanks! I will make these changes.

>> I think these datasets are often consumed/prepared by people who have no idea/interest in what they're processing or how they are actually used in DD

I think if the user does not want to use DD to process this data, we do provide the CoNLL format, which is a standard format that user can use.

>> In fact, the opendata TSVs are slightly different from the walkthrough: they have provenance and word index columns, sentence id appears after document id, the dependency paths are represented differently using two columns. 

This might not be a short-term thing, but I feel like we need to change the walkthrough to make it Bazaar-compliant OR make Bazaar walkthrough-compliant. Open data is not the place that we define the schema; Bazaar is the place that we define the schema for both open data and walkthrough to use.

zhangce:
@netj I did another pass.

netj:
Looks great!

On the first point, I meant there may be professional DBAs involved who prepares the database for the actual DD user, who don't necessarily want to follow the details of DD tutorial.  Anyway the new schema page looks great.

I completely agree the tutorial should be updated at one point to include Bazaar and embrace its format.

PR #415, opened 2015-09-26
add back missing double quotes in deepdive.conf

Atlas7:
I believe the pair of double quotes might be missing in this deepdive.conf - please kindly check (and many thanks for the great repository!)


chrismre:
Thanks for the kind words! @netj @zhangce ?

netj:
Thanks for your fix!  deepdive.conf is in [HOCON syntax, which doesn't require double quotes](https://github.com/typesafehub/config/blob/master/HOCON.md#unquoted-strings), but it's nice to have them for consistency within that example.

Atlas7:
@netj oh I see - thanks and keep up the great work!

PR #417, opened 2015-09-28
Fix bug in specifying deepdive output folder using deepdive run -o

feiranwang:
The bug is: when specifying a folder with -o, `deepdive-run` prepends an extra path to app folder.


netj:
Thanks for fixing this bug.  Merging with a minor bash style fix.

Issue #411, opened 2015-09-25
Cannot get features for mindtagger


raphaelhoffmann:
With DeepDive's new schemas, I am no longer able to display features and feature weights in mindtagger.

Our app.ddlog contains multiple extractors. With only a single extractor, I can do:
```
t1=TABLENAME
t2=inference_rule_0
psql DBNAME -c "
         SELECT \"${table}_inference.R0.id\"
              , ARRAY_AGG(\"dd_weight_column_0\" ORDER BY abs(weight) DESC) AS features
              , ARRAY_AGG(weight  ORDER BY abs(weight) DESC) AS weights
           FROM dd_query_$t2 f
              , dd_inference_result_weights_mapping wm
          WHERE wm.description = ('$t2-' || \"dd_weight_column_0\")
          GROUP BY \"${table}_inference.R0.id\"
"
```

However, with multiple extractors, the relevant table may be named ```dd_query_inference_rule_0```, ```dd_query_inference_rule_1```, ```dd_query_inference_rule_2```, ..., and I don't know how to select the right one.

PR #418, opened 2015-09-28
Incremental update

feiranwang:
1. Update incremental doc, fixing typos etc.
2. Update incremental spouse example, reflecting the latest 
3. Update scripts for running incremental spouse example, using deepdive commands


.gitignore
```diff
@@ -30,3 +30,4 @@
 .mindtagger.conf.json
 /out
 /log
+/lib
```
netj: Why is this added back?

examples/spouse_example/postgres/incremental/setup_database.sh
```diff
@@ -12,9 +12,11 @@ fi
...
+# dropdb $DBNAME || true
+# createdb $DBNAME
 
-psql -d $DBNAME <./schema.sql
+# psql -d $DBNAME <./schema.sql
 bzcat ../../data/articles_dump.csv.bz2  | psql -d $DBNAME -c "COPY articles  FROM STDIN CSV"
-bzcat ../../data/sentences_dump.csv.bz2 | psql -d $DBNAME -c "COPY sentences FROM STDIN CSV"
+bzcat ../../data/sentences_dump.csv.bz2 |
+if [[ -z ${INCREMENTAL_SPOUSE_EXAMPLE_LIMIT_SENTENCES:-} ]]; then cat; else head -n ${INCREMENTAL_SPOUSE_EXAMPLE_LIMIT_SENTENCES}; fi |
+psql -d $DBNAME -c "COPY sentences FROM STDIN CSV"
```
netj: Let's use `deepdive sql`.

examples/spouse_example/postgres/incremental/spouse_example.f1.active.rules
```diff
@@ -1,3 +1,3 @@
 # List names of the inference rule to mark as active in materialization phase here.
 # It's in the form of the relation name of the head in the inference rule followed by an underscore and the zero-based index of the rule.
-has_spouse_0  # You can put comments after a hash letter.
+inference_rule_0  # You can put comments after a hash letter.
```
netj: This is another place where the compiler naming convention needs to be fixed as pointed out in #411.

shell/deepdive-run
```diff
@@ -73,6 +73,7 @@ else
     run_dir="$DEEPDIVE_APP/run/$run_id"
 fi
 mkdir -p "$run_dir"
+mkdir -p run
```
netj: Is this necessary?  If so, let's add a comment.

.gitignore
feiranwang: I was thinking we don't want to track the library files. `library/` is also not included in the source tree, so we are currently not tracking it.

netj:
Documentation update looks good.
A few high-level comments:
* Ideally, we should get rid of run.sh, and have `deepdive run` accept a phase (mode) option that gets passed down to `ddlog compile`.
* Most of setup_database.sh lines should be moved to input/init.sh, and `deepdive initdb` should be used instead of `createdb` in 0-setup.sh.

.gitignore
netj: But do any of our source or build produce `lib/` or `library/`?  I don't think we need to over ignore.

feiranwang: I see. It's probably old files on my laptop. I'll remove this.

feiranwang:
Updated and rebased so that the `.gitignore` change is not included.

PR #435, opened 2015-10-26
adds an environment variable for INPUT_BATCH_SIZE

xiaoling:
cf HazyResearch/ddlog@384488ceb6a5c3ac8bd5a9a96c8c5861be366ffa

### DIFF
--- a/shell/deepdive-run
+++ b/shell/deepdive-run
@@ -110,7 +110,7 @@ fullConfig=$run_dir/deepdive.conf
     export PARALLELISM
     
     # set INPUT_BATCH_SIZE env var 
-    export INPUT_BATCH_SIZE=10000
+    ! [[ -e "$INPUT_BATCH_SIZE" ]] || export INPUT_BATCH_SIZE=10000
 }
 
 # user's deepdive.conf overrides anything compiled from DDlog
--- a/shell/deepdive-run
+++ b/shell/deepdive-run
@@ -108,6 +108,9 @@ fullConfig=$run_dir/deepdive.conf
         echo 1
     ) 2>/dev/null)}
     export PARALLELISM
+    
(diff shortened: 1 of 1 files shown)


shell/deepdive-run
```diff
@@ -108,6 +108,9 @@ fullConfig=$run_dir/deepdive.conf
         echo 1
     ) 2>/dev/null)}
     export PARALLELISM
+    
+    # set INPUT_BATCH_SIZE env var 
+    export INPUT_BATCH_SIZE=10000
```
netj: Always settings this prevents the user from overriding the value.  Please fix it to respect user's environment.

shell/deepdive-run
```diff
@@ -108,6 +108,9 @@ fullConfig=$run_dir/deepdive.conf
         echo 1
     ) 2>/dev/null)}
     export PARALLELISM
+    
+    # set INPUT_BATCH_SIZE env var 
+    ! [[ -e "$INPUT_BATCH_SIZE" ]] || export INPUT_BATCH_SIZE=10000
```
netj: First, when `INPUT_BATCH_SIZE` is not defined, the entire script aborts.  Second, this test with `-e` is checking whether there *exists* a file named with that value which doesn't make sense.

Here's the right way to do this:
```bash
: ${INPUT_BATCH_SIZE:=10000}
export INPUT_BATCH_SIZE
```

xiaoling: You're right. Sorry my bash is a bit rusty. I should've been more careful.

netj: No worries.  Can you also update the ddlog submodule to point to the commit relevant to this?

netj:
LGTM

PR #437, opened 2015-11-01
Speed up grounding performance

SenWu:
1. Get rid of factor id assignment in grounding part when Deepdive runs original mode.
2. Keep everything the same in incremental mode.


src/main/scala/org/deepdive/inference/Sampler.scala
```diff
@@ -79,7 +79,6 @@ class Sampler extends Actor with ActorLogging {
           "-w", weightsFile,
           "-v", variablesFile,
           "-f", factorsFile,
-          "-e", edgesFile,
```
netj: We should probably update the sampler documentation as well.  Could you open an issue for that so we don't forget?

netj:
Looks good.  Now the grounding code looks even more scarier, but it'll hopefully get ironed it out as we migrate to the new data flow compiler architecture.

src/main/scala/org/deepdive/inference/Sampler.scala
SenWu: Sure, let me do it.

PR #438, opened 2015-11-04
Add three-value imply function

feiranwang:


netj:
Does the sampler already support the Imply3 factor or did you forget to add the submodule changes?

feiranwang:
Yes, the sampler already has support for this. https://github.com/HazyResearch/sampler/blob/master/src/dstruct/factor_graph/factor.h#L151

netj:
LGTM

PR #440, opened 2015-11-18
Remove linux sampler shared library zip file. Copy necessary libraries at build time

feiranwang:


feiranwang:
Updated. PTAL.

netj:
Awesome!

Issue #426, opened 2015-10-02
Parallelization


vsoch:
I have my extractors working, SSL resolved (thanks @netj !) and I've been started to learn about Hadoop getting ready to figure out Parallelization. Then I came across [this paper](http://www.cs.stanford.edu/people/chrismre/papers/deepdive_vlds.pdf).

 >> We use the Hadoop File System for storage, but found a 100-node MapReduce cluster to be insufficient for ClueWeb: (1) Hadoop’s all-or-nothing approach to failure handling hinders throughput, and (2) the number of cluster machines for Hadoop is limited

Before I dive too deeply (pun intended) into learning an infrastructure that has been tested and deemed not appropriate, I wanted to run this by you again. The clusters that are available are:

#### TACC 
(specifically, Wrangler, where I am working now, which has `Hadoop` reservations, `SLURM` (with a limited job queue), a `"Launch"` script that will take a list of single commands as input and run them all at once on a grid system (a long wait time, but pretty diabolical and quick when you get access, and only prone to error if there are issues with accessing / reading writing the same files).

### Sherlock
Is here at Stanford, and it's a standard `SLURM` queue with 5000 jobs able to queue, and we have about 25 private nodes that, although it's not a super TACC infrastructure, I haven't hit anything yet that I haven't been able to script up in some parallel fashion, and either let run or manage jobs over a week to make sure it's all done properly.

My instinct is telling me that, most people are going to have access to SLURM, and even though it's not "click one button and go away," a pipeline could be set up that is logical, clear, and workable.  My curiosity is urging me to give Hadoop a try, and minimally, learn lessons the hard way. My logical mind is telling me that this question is likely been asked many times before, I should go straight to the source to pursue the solution that is most promising, hence why I am writing this post :_)

PR #445, opened 2015-12-17
DeepDive execution plan compiler

### LINKED ISSUE #1: Configuration sanity check
We should do a sanity check on the configuration upon loading. Instead of the application crashing in the middle of execution due to a configuration issue, we should immediately exit if we find an obvious mistake. Some things we can check for:

- Does the schema match the variables used in factor functions?
- Do all extractor UDF files exist?
- Do all extractor output relations exist?

There probably are more things that we can check for.

### LINKED ISSUE #20: Catch the exception of no database connection
There is no exception catch for database connection error, and the program runs forever...

### LINKED ISSUE #110: enable SQL comments
Users cannot use comments for SQL queries in current system. This is because the system will simply joining all lines (which might make the comment commenting all lines)

These include:
1. sql_extractor.sql
2. extractors.input
3. factors.input_query

We should enable users to use comments somehow..

### LINKED ISSUE #273: extractorRunner hanging problem
I am using develop branch:

```
15:43:10 [DataLoader(akka://deepdive)] INFO  gpload -f /tmp/gpload9045506308201380182.yml
15:43:10 [Helpers$(akka://deepdive)] INFO  Executing command: "/tmp/gpload5879879319714338796.sh"
15:43:10 [Helpers$(akka://deepdive)] INFO  /tmp/gpload5879879319714338796.sh: line 1: gpload: command not f
ound
15:43:10 [extractorRunner-ext_mention] ERROR Failure when executing script: /tmp/gpload58798793197[48/1392]
sh
java.lang.RuntimeException: Failure when executing script: /tmp/gpload5879879319714338796.sh
        at org.deepdive.helpers.Helpers$.executeCmd(Helpers.scala:102) ~[deepdive_2.10-0.0.3.jar:0.0.3]
        at org.deepdive.datastore.DataLoader.load(Dataloader.scala:141) ~[deepdive_2.10-0.0.3.jar:0.0.3]
        at org.deepdive.extraction.ExtractorRunner.org$deepdive$extraction$ExtractorRunner$$runTsvExtractor
(ExtractorRunner.scala:492) ~[deepdive_2.10-0.0.3.jar:0.0.3]
        at org.deepdive.extraction.ExtractorRunner$$anonfun$2.applyOrElse(ExtractorRunner.scala:131) ~[deep
dive_2.10-0.0.3.jar:0.0.3]
        at org.deepdive.extraction.ExtractorRunner$$anonfun$2.applyOrElse(ExtractorRunner.scala:100) ~[deep
dive_2.10-0.0.3.jar:0.0.3]
        at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:33) ~[scala-library-2.
10.5.jar:na]
        at akka.actor.FSM$class.processEvent(FSM.scala:587) ~[akka-actor_2.10-2.3-M2.jar:2.3-M2]
        at org.deepdive.extraction.ExtractorRunner.processEvent(ExtractorRunner.scala:59) ~[deepdive_2.10-0
.0.3.jar:0.0.3]
        at akka.actor.FSM$class.akka$actor$FSM$$processMsg(FSM.scala:581) ~[akka-actor_2.10-2.3-M2.jar:2.3-
M2]
        at akka.actor.FSM$$anonfun$receive$1.applyOrElse(FSM.scala:575) ~[akka-actor_2.10-2.3-M2.jar:2.3-M2
]
        at akka.actor.Actor$class.aroundReceive(Actor.scala:467) ~[akka-actor_2.10-2.3-M2.jar:2.3-M2]
        at org.deepdive.extraction.ExtractorRunner.aroundReceive(ExtractorRunner.scala:59) ~[deepdive_2.10-
0.0.3.jar:0.0.3]
        at akka.actor.ActorCell.receiveMessage(ActorCell.scala:491) [akka-actor_2.10-2.3-M2.jar:2.3-M2]
        at akka.actor.ActorCell.invoke(ActorCell.scala:462) [akka-actor_2.10-2.3-M2.jar:2.3-M2]
        at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:237) [akka-actor_2.10-2.3-M2.jar:2.3-M2]
        at akka.dispatch.Mailbox.run(Mailbox.scala:219) [akka-actor_2.10-2.3-M2.jar:2.3-M2]
        at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(AbstractDispatcher.scala:385) [
akka-actor_2.10-2.3-M2.jar:2.3-M2]
        at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260) [scala-library-2.10.5.jar:n
a]
        at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339) [scala-library-
2.10.5.jar:na]
        at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979) [scala-library-2.10.5.j
ar:na]
        at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107) [scala-library
-2.10.5.jar:na]
15:43:10 [extractorRunner-ext_mention] INFO  waiting for tasks
15:43:10 [extractorRunner-ext_mention] WARN  unhandled event PrintStatus in state Idle
15:43:40 [extractorRunner-ext_mention] WARN  unhandled event PrintStatus in state Idle
```

Then it hangs there. I should have fixed it before... need to check.

### LINKED ISSUE #280: Restart DeepDive from a "checkpoint"
Currently when DeepDive quits during grounding or sampling, we do not have the ability to recover from a near point the next time it runs. The closest thing we can do is restarting the whole grounding, or using "relearn_from" parameter to restart from the grounded factor graph. 

Following features should make recovering easier:

1. If DeepDive quits during grounding, next time it executes it can restart from the query in grounding where it quits.
2. If DeepDive quits anywhere after sampling, it should be able to start from loading the sampler results, instead of rerunning the sampler.

### LINKED ISSUE #329: Refactor Scala code

### LINKED ISSUE #349: deepdive run doesn't report an error when extractor dependencies are misspelled
It's more or less a silent failure.  The only indication in the logs is order in which the extraction manager executes the extractors.

### LINKED ISSUE #351: Separate initdb command / flag for input data vs. extractions schemas
This is minor / low-priority, but it's convenient to be able to e.g. add a new entity type or otherwise tweak the schema, and then run something like initdb, without also dropping the sentences table say...

### LINKED ISSUE #357: schema.sql file generated from ddlog?
Is this on the roadmap out of curiosity?  Seems like this would be nice and also fairly simple way to centralize the basic DD app in the ddlog file- or is there a way to do this already?  Thanks!

### LINKED ISSUE #361: INSERT count update error with Greenplum
Also with psql < 9.3

See [this post](http://stackoverflow.com/questions/27043186/org-postgresql-util-psqlexception-unable-to-interpret-the-update-count-in-comma)

[This post](https://www.pivotalguru.com/?p=414) says that the data actually gets updated, we just need to catch this error and run an `ANALYZE TABLE` query...

### LINKED ISSUE #383: Error messages not appearing in log file
Some error messages only appear on the console, but not in the log file, eg. #381.

### LINKED ISSUE #384: PGXL PL/Python extractors keep running after Ctrl-C'ing DeepDive
Not sure if we can do something about this. Hoping we'll eventually get the same speed with tsv_extractor and can avoid PL/Python.

### LINKED ISSUE #412: deepdive run with ddlog is not idempotent
Suppose you have a ddlog program with the following structure:
```
table1(
...
)
function f over () returns ... implementation ...
table1 += f(...) :- ...
```

Running ```deepdive run``` multiple times is not idempotent on such programs. More rows will be appended to table1 in each run.

Is there a way to overwrite or clear the tables? Using '=' instead of '+=' throws a compiler error.

### LINKED ISSUE #421: DDLOG fails when views depend on views
Consider the following DDLOG snippet:
```
wordidxs_unnested(doc_id, sent_id, UNNEST(wordidxs)) :-
  locations(id, doc_id, sent_id, wordidxs, mention_id, type, entity, words, is_correct, features).

wordidxs_tmp1(doc_id, sent_id, MIN(wordidxs) - 1) :-
  wordidxs_unnested(doc_id, sent_id, wordidxs).
```
Running this program the second time crashes DeepDive, because it cannot drop view ``wordidxs_unnested`` (view ``wordidxs_tmp1`` depends on it).

There may be a better fix, but one work-around is to generate ``deepdive.conf`` and append ``CASCADE`` to
```
DROP VIEW IF EXISTS wordidxs_unnested;
```

### LINKED ISSUE #427: If an extractor depends on inference, DD halts indefinitely before doing inference
At least stop with an error ... please. What's currently happening is that the output looks exactly like inference (with lines like

    09:10:33 [taskManager] INFO  Memory usage: 539/1809MB (max: 233017MB)

getting output every minute and nobody knows what's going on.

(ACTUALLY, I'd propose ALLOWING extractors after inference. MY particular extractor was for printing precision and recall numbers after every run. Prolepsis: This SHOULD be in Dashboard, and it's going to be there, but it's just nice to have a precision number immediately after a run, in the DD log output.)

### LINKED ISSUE #428: Possible to get rid of UDF input data splitting?
We ran a simple extractor that basically just takes phrases with the NER tag PERSON over 30M documents (about 200GB in terms of sentences table size). Here is a breakdown of the time spent:

COPY OUT (with ARRAY_TO_STRING): 30 min
split: 35 min
UDF: 10 min (70 threads)
COPY IN: 20 min

First, some small perf comments:
1. Wonder whether we could speed up COPY OUT significantly if we could get rid of ARRAY_TO_STRING one day...
2. DD by default seems to split the UDF input data into small files with 10K lines each. The DDLog syntax doesn't seem to support the specification of the UDF input batch size. Following @SenWu 's suggestion, we used a larger batch size (100K lines), and `split` ended up taking 35 min. Not sure if it would be much faster if we use an even larger batch size. Small batch sizes have two issues: a) too many files making `split` itself slow; b) if the UDF has a slow init step, that init overhead is multiplied by the number of splits. Of course, if the batch size is too large, and the number of splits << number of cores, we under-parallelize. Maybe we could do some adaptive splitting; e.g., split into 100MB chunks, and if it ends up having too few chunks, further split each chunk so that the total number of chunks ~= number of cores.

And the big comment:

Is it possible to get rid of the file splitting step altogether? If we could pipe COPY OUT | UDF | COPY IN all in memory, we might be able to run the above program in 20 min instead of 100 min. ( @netj had some great hacks that seem to be in this direction https://github.com/netj/mkmimo . Has that been applied in DD anywhere?)

### LINKED ISSUE #431: Why remove tasks dependencies that are not in the pipeline?
What is the rationale for removing task dependencies not specified in the pipeline?
https://github.com/HazyResearch/deepdive/blob/master/src/main/scala/org/deepdive/DeepDive.scala#L148

By always removing the dependencies one is forced to resolve the dependency graph manually and add to the pipeline.  I do see the benefit of allowing the pipeline definition to override dependencies but it seems this should not be the default behavior.

I may create a PR to add a runtime arg to control this... I am eager to get in a commit!

### LINKED ISSUE #441: no install instructions for building from source
so after i git clone this repo, then what?

http://deepdive.stanford.edu/doc/basics/installation.html

that page says nothing about doing a git clone and how to build from scratch.

netj:
This not-so-small PR adds a new implementation of DeepDive that literally compiles an *execution plan* to run the app in a much more efficient way in both human time and machine time/space.  I invite everyone to review the code and give a try on your existing app (especially @feiranwang @SenWu @ajratner @zhangce @raphaelhoffmann @alldefector @zifeishan @xiaoling @Colossus @ThomasPalomares @juhanaka)  The plan is to get feedback over the next few days while I update the documentation and carve out release v0.8.0.

An execution plan is basically a set of shell scripts that tells what to run for user-defined extractors as well as the built-in processes for grounding the factor graph and performing learning and inference, and a Makefile and that describes the complete dependencies among them.  These are compiled from the app's deepdive.conf, app.ddlog, and schema.json, mainly by a series of JSON transformations implemented as [jq programs](https://stedolan.github.io/jq/tutorial/) with a little bit of help from bash, which all resides under `compiler/` in the source tree.

It implements most of the existing functionality provided by the current Scala implementation, except a few things, e.g., Greenplum parallel unloading/loading, which won't be difficult to add given the much much modular architecture.  On the other hand, exciting new features and improvements are added.  To highlight just a few:
* **Full Dependency Support with Selective Execution.** It's now possible to selectively run, repeat, or skip certain parts of the app's data flow or extraction pipeline without being aware of all the dependencies between them. (fixes #431, closes #427, closes #273)  The user has full control over every step of the execution plan.  Not only that, but grounding is also broken down into smaller processes, so it's possible to just change or add one inference rule and update the grounded factor graph without having to recompute everything from scratch. (fixes #280)
* **Zero Footprint Extraction.** `tsv_extractor`s now have nearly zero footprint on the filesystem.  The data is streamed from the database through the UDFs and back to the database.  [mkmimo](https://github.com/netj/mkmimo) is used with named pipes to make the connection between the database and the parallel UDF processes (cf. `$DEEPDIVE_NUM_PROCESSES`). (fixes #428)  It doesn't support other extractor styles yet, but we can probably drop them unless there's a compelling reason. (closes #384)
* **Compute Drivers.** A new compute driver architecture for executing such extractors is now in place, so it's now clear where to extend to support remote execution or clusters with a job scheduler, such as Hadoop/YARN, SLURM, Torque/GridEngine/PBS. (#426)  The local execution driver is what implements the streaming `tsv_extractor` mentioned above.  Moreover, the grounding processes as well as the user-defined extractors also make use of the compute drivers, so some parts of the grounding will automatically take advantage of such extensions.
* **Zero Footprint Grounding.** The grounding processes also minimizes footprint on the filesystem.  No data for the factor graph are duplicated.  Instead of creating a concatenated copies of factors, weights, and variables, they are merged as the sampler loads them.  Also, the binary format conversion is done on the fly as the grounded rows are unloaded from the database, so no ephemeral textual form is ever stored anywhere.  In fact, only a few line changes to the compiler can compress the binary forms and shrink the factor graphs footprint on the filesystem by an order of magnitude (not included in this PR).
* **More User Commands.** `deepdive.pipeline` config is obsolete as well as the `deepdive run` command, although they'll work the same as before.  Now, the user can simply state what is the goal for execution to the `deepdive do` command, e.g., `deepdive do model/calibration-plot` or `deepdive do data/has_spouse_features`, as many times as necessary once the app is compiled with the `deepdive compile` command.  Supporting commands such as `deepdive plan`, `deepdive mark`, `deepdive redo` are there to speed up typical user workflows with DeepDive apps.  `deepdive initdb` and `deepdive load` have been improved to be more useful, and is used by a few compiled processes. (fixes #351, fixes #357)
* **Error Checking.** Errors in the app (and of course the compiler itself) are checked at compile time, and can also be done by the user with the `deepdive check` command.  The checkers are modular and isolated, so many useful checks can be quickly added, such as test firing UDF commands, etc.  Only basic checks have been implemented so far. (fixes #349, fixes #1)
* **Simpler, Efficient Multinomial Factors.** Multinomial factors are won't materialize unnecessary data and use VIEWs as much as possible, e.g., dd_*_cardinality tables and the dd_graph_weights.  Also, nearly no code has been duplicated to support it, so that's good news for developers.
* **Bundled Runtime Dependencies.** DeepDive now builds essential runtime dependencies, such as bash, coreutils, jq, bc, graphviz, so no more Mac vs. Linux or software installation version issues will pop up. (fixes #441 as documentation also ended up in this PR)

Also some good stuffs that comes with a clean rewrite:
* Closes #412 by doing a reasonable job and making the base relations clear to the user, and also closes  #421.
* Closes #110 as no more SQL parsing is done.
* Closes #383 as logging has been completely redone.
* Closes #361, closes #20 as JDBC is no longer used.
* Closes #329 as Scala implementation will be simply dropped in a future release (v0.8.1 or v0.9?).


chrismre:
Wow.

Colossus:
That's crazy

alldefector:
Sounds like 1.0 instead of 0.8 :)

chrismre:
Let's not get crazy.

netj:
FYI Here's the SVG rendered [data flow for spouse example](https://gist.github.com/netj/f4abe6eff63b837c7046#file-dataflow-svg) and [data flow for chunking example](https://gist.github.com/netj/a4e363b92ffeb2fa95b7#file-dataflow-svg).

I forgot to mention that this PR requires **no change to the user's app code**, except maybe a few extra dependency information that were missing in the manually written deepdive.conf.  The `run.sh` that typically handled env setup and many other hacks should be moved to the right place such as input/.  @ThomasPalomares and I were able to get the genomics app (in DDlog) working with minimal changes.

chrismre:
This is awesome :)

runner/compute-driver/local/compute-execute
```diff
@@ -0,0 +1,136 @@
...
+
+    # show configuration
+    echo " output_relation_tmp=$output_relation_tmp"
+    echo
+
+    # use an empty temporary table as a sink instead of TRUNCATE'ing the output_relation
+    deepdive-initdb "$output_relation"
+    deepdive-sql "
+        DROP TABLE IF EXISTS $output_relation_tmp;
+        CREATE TABLE $output_relation_tmp (LIKE $output_relation);
```
alldefector: Maybe add `INCLUDING ALL` to make sure all other bells and whistles are retained?

runner/compute-driver/local/compute-execute
```diff
@@ -0,0 +1,136 @@
...
+    # use mkmimo to distribute input data to multiple processes
+    mkmimo /dev/stdin \> process-*.input &
+fi
+
+wait  # until everything is done ##############################################
+
+if [[ -n $output_relation ]]; then
+    # rename the new temporary table
+    # TODO maybe use PostgreSQL's schema support here?
+    echo "Replacing $output_relation with $output_relation_tmp"
```
alldefector: Is inserting into (as opposed to replacing) the target table something we need?

runner/compute-driver/local/compute-execute
netj: Agreed and done.  Thanks for suggesting this.

runner/compute-driver/local/compute-execute
netj: I'd say no for running UDFs, as it makes part of the data flow non-idempotent, hence unwieldy for the user as well as the system.  Whenever a user needs to append the output of some UDF to somewhere or to take a UNION of multiple UDFs, sql_extractor or cmd_extractor is there.  DDlog compiler may need to be adjusted to automatically compile a UNION VIEW when multiple UDFs define the same head.  Related to this, I added a check that ensures all output_relation is defined by a single extractor, as it'll surprise all users when multiple extractors are defined to output to the same relation but only one extractor's output remains.

compiler/compile-code/compile-code-piggy_extractor
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env jq-f
+# compile-code-plpy_extractor -- TODO Compiles plpy_extractors in the normalized JSON into code executable by the runner
+##
+.deepdive_ as $deepdive
+
+| $deepdive.execution.processes | to_entries[]
+| .value.name = .key | .value
+| select(.style == "plpy_extractor")
```
zifeishan: It seems that the content in `compiler/compile-code/compile-code-piggy_extractor` and `compiler/compile-code/compile-code-plpy_extractor` should switch?

netj: Good catch!  I'm actually planning to drop these in a later commit: f5186538f47e5a1f5020ce2420f3ec49b30a5b11 so maybe it's not too critical.

Makefile
```diff
@@ -130,14 +137,17 @@ util/format_converter_mac: src/main/c/binarize.cpp
 endif
 
 .PHONY: build-mindbender
+MINDBENDER=mindbender
 build-mindbender:
-	git submodule update --init mindbender
-	$(MAKE) -C mindbender clean-packages
-	$(MAKE) -C mindbender package
+	@util/build/build-submodule-if-needed $(MINDBENDER) mindbender-LATEST-Darwin-x86_64.sh
```
zifeishan: Seems missing `test-build build: build-mindbender` here

compiler/compile-config/compile-config-2.01-grounding
```diff
@@ -0,0 +1,739 @@
...
+        table=\(.variablesTable | @sh)
+
+        cd \"$DEEPDIVE_GROUNDING_DIR\"/variable/${table}
+        baseId=$(cat id_begin)
+
+        # assign id to all rows according to the paritition
+        deepdive db assign_sequential_id $table \($deepdiveVariableIdColumn | @sh) $baseId
+
+        \(
+        if .variableType == "categorical" then
```
feiranwang: The "categorical" branch seems like a separate part from id assignment.

Makefile
netj: mindbender build has been optional in the past as it's quite a big one, so I kept it as it is.  (stage.sh also optionally stages the built binary.)  I feel we should do that in a separate PR

compiler/compile-config/compile-config-2.01-grounding
netj: I agree.  Maybe we should rename this step to a more general one?  Any good idea?  Alternatively, we could create optional process nodes, but I didn't want to mess up with the dependency graph as it requires another branch when enumerating dependencies for the `"process/grounding/variable_holdout"` node.

runner/compute-driver/local/compute-execute
alldefector: Makes sense. Though the `+=` syntax in ddlog may give people the false impression that the output would be appended...

zifeishan:
The `deepdive sql '[multiple SQL queries]'` does not work under postgres-XL. It's probably necessary to manually segment these queries into individual ones when compiling the run scripts.

Makefile
zifeishan: Also, why are we forcing the `Darwin` release here?

raphaelhoffmann:
Hi @netj, this new version of deepdive is really great! Here are a few notes I took while experimenting.

1. When building on OSX, I got error that X11/Xlib.h couldn't be found. This error
was thrown by buildkit module for graphviz.
    ```
    /usr/include/tk.h:78:23: fatal error: X11/Xlib.h: No such file or directory
    ```
    Following [these](http://stackoverflow.com/questions/11465258/xlib-h-not-found-when-building-graphviz-on-mac-os-x-10-8-mountain-lion) instructions, I fixed it by adding a symlink
    ```
    ln -s /opt/X11/include/X11 /usr/local/include/X11
    ```
    Afterwards, it was not sufficient to rerun 'make'. I manually cd'ed to ```deepdive/depends/bundled/graphviz/graphviz-2.38.0``` and ran ```./configure && make && make install```. Then I could run "make" in deepdive

2. ddlog apps didn't work with the default submodule commit, but – following your
   suggestions – everything worked fine after upgrading to ddlog commit ee0b359c9f018c19df5eddbf3bff60bdf076e091.

Now everything works fine, but I've encountered two different problems in a small number of runs:

1. Sometimes, stdin is no longer showing on the screen after a successful ```deepdive do```
   execution. That is, I can type commands and they execute, but the letters I type are not
   being shown on the screen. I'm using iterm2.

2. Sometimes (rarely), a mkmimo process kept running at 100% cpu usage, after ```deepdive do``` 
   terminated with an error (possibly related to missing input data).

I'll try to investigate more carefully if I encounter these again.

chrismre:
Great feedback!!

runner/compute-driver/local/compute-execute
netj: Would it be clearer if multiple `+=` ddlog rules to the same head relation were compiled into a union of each rule's output?  That way, the semantics still is augmenting the relation with more tuples from the function, but beginning from an empty relation for a given program. Currently, having such repeated heads will result in a deepdive compiler error.

Makefile
netj: That's a mistake. Fix is on the way :)

netj:
@zifeishan Regarding multiple SQL statements, I'm reluctant to do a blind split on semicolons, as any string literal may contain it.  Do you know why PGXL is not supporting them or any workaround?  One safe way for DD to deal with multiple statements is to support an array of strings for the sql field for sql_extractors, although this will break backward compatibility (but who cares! : )  How does it sound?

@raphaelhoffmann Thanks for your detailed feedback!

1. The runtime dependency build hasn't been thoroughly tested, hence such bugs.  Specifically on graphviz, I feel the build can be simplified without pulling in X11.  Will come up with a fix.
2. The submodule build has been modified to honor the developer's working tree, so whatever is on disk gets built, as opposed to forcing all submodules to be the exact version recorded in deepdive's commit or the git index.  So forgetting to run `git submodule update --init` before a build can now result in an unexpected result.  Previously, running build without a `git add path/to/submodule` wiped out changes to the submodule, upsetting people who were modifying them.  The previous behavior seems to be upsetting less number of people, so I'll revert that or maybe abort the build instead of silently making surprises under the hood.
3. The terminal messup is due to the pv command showing progress bar, competing with the logs printed to the same terminal.  I tried to reduce that error but haven't figured out how to completely eliminate that.  As a workaround you can run `reset` whenever that happens by blinding typing it followed by an return/enter.  We need a tty expert here.  There should be a simple invisible escape sequence we can print to ensure the terminal is restore in a sane state.
4. mkmimo had a busy waiting bug on OS X where the poll(2) syscall returns too fast with slow sinks.  I added a fix for it in f2fabb16cd5b6efab237bb7bf9ec29c97971d851, but kept it separate to keep this giant PR stable for review for a while.  Since we've got some eyes and hands over the code, I can bring in the latest changes, but that'll require another round of tests to be safe.

runner/compute-driver/local/compute-execute
alldefector: That sounds great :)

zifeishan:
@netj 
This seems to be a bigger issue than `sql_extractor` for PGXL: I am trying out some other extractors but it had other problems with this line: https://github.com/HazyResearch/deepdive/blob/dataflow-compiler/runner/compute-driver/local/compute-execute#L123 since it's also calling multiple queries with `deepdive sql`. I guess a cleaner workaround would be splitting queries for  `https://github.com/HazyResearch/deepdive/blob/dataflow-compiler/database/db-driver/postgresql-xl/db-execute`?

Regarding the PGXL issue, I could not find discussions for it online. A sample query that would fail on my environment:

```
$ psql $DBNAME -c  "create table tmptmp(id int);
insert into tmptmp values(1);
select * from tmptmp;
drop table tmptmp;"
ERROR:  Unexpected response from the Datanodes for 'T' message, current request type 1
ERROR:  Unexpected response from the Datanodes for 'T' message, current request type 1
ERROR:  Unexpected response from the data nodes for 'D' message, current request type 0
```

Server version: Postgres-XL 9.2.0.

netj:
@zifeishan If this is a blocker, we can quickly apply the workaround to pgxl driver, which is actually super simple.  `db-execute` can call `psql` multiple times after splitting the query argument by semicolon and that'll handle everything.  Wanna take a stab at it?

zifeishan:
@netj I just found an alternative and it works: instead of 
```
deepdive sql  "create table tmptmp(id int);
insert into tmptmp values(1);
select * from tmptmp;
drop table tmptmp;"
```
, if I run:
```
echo  "create table tmptmp(id int);
insert into tmptmp values(1);
select * from tmptmp;
drop table tmptmp;" | deepdive sql
```
it will work for PGXL.

netj:
@zifeishan Interesting.  However, that'll take away stdin for psql.  Is there a way to tell psql a sql file to run instead?

zifeishan:
@netj Yes. Try `psql -f commands.sql`. It works with pgxl as well.

alldefector:
@zifeishan @netj are you guys suggesting that we write `$sql` to a temp file and use `-f` to run it (so that the whole block isn't run as one transaction)? Sounds like it should work. We only need to update PGXL's db-execute driver, right?

netj:
@alldefector Sorry the conversation went private through Slack.  Yes, that's exactly our plan, except using bash processes substitution to turn `$sql` into a readable file.  This is so much better than dangerously splitting the SQL queries in a sloppy way.  (btw you mean the block *is* run as a transaction, right?)

@zifeishan Could you make the changes, confirm it working, and add the commit to this PR?

alldefector:
Apparently `-f` does not run the file in one transaction unless you also specify `-1` or `--single-transaction`: http://postgres-xc.sourceforge.net/docs/1_0/app-psql.html

That's actually what we need because PGXL would throw up (Zifei's error messages) if we try to run a bunch of mutating / DDL statements in one transaction.

feiranwang:
@netj I tried this on several examples. `deepdive run` and  `deepdive do all` work great. I have a small problem with `deepdive do`. When I type `deepdive do`, it gives a list of targets. In smoke example there's a `process/grounding/variable/person_has_cancer/assign_id`. When I tried to run with that target, it gives an error
```
process/grounding/variable/person_has_cancer/assign_id: Unknown target
```

netj:
@feiranwang Can you try putting a `.done` suffix to the target? Maybe I forgot to add that case. Will also take a look into this.

feiranwang:
@netj  I added a `.done` suffix to the target and it gives the same error
```
process/grounding/variable/person_has_cancer/assign_id.done: Unknown target
```

netj:
@feiranwang It turns out to be a missing schema.json issue, so I converted schema.sql into app.ddlog as an easy way to generate the json file.  The compiled Makefile was missing some dependencies for the mentioned target, which gets generated from the relational schema.  I fixed things to show such errors more transparently.  Ideally, these should be caught in the `deepdive compile` phase.  It'd be nice if you can contribute a checker that prevents this kind of error.  Also, it seems we're missing a test for the smoke example.

PR #446, opened 2015-12-29
Fixed weight table for Postgres-XL

zifeishan:
Postgres XL cannot do distributed update over distribution key. Fixed the grounding queries to support `weight_table` with Postgres-XL.


alldefector:
With the change, you are still updating the `initvalue` field; and it's not a key column.

Real problem seems to be that PGXL doesn't support correlated UPDATE involving multiple nodes:
http://sourceforge.net/p/postgres-xc/mailman/message/26727433/

There is a newer version of PGXC (basis of PGXL) that may have solved that limitation:
https://github.com/postgres-x2/postgres-x2
We could try it at some point.

Anyway, I heard that the scala code will go away right after the 0.8 release?

netj:
Regarding the future of Scala code, I'm still deciding whether to drop it even earlier and keep the latest in something like 0.7.2, or to have it coexist with the compiler in 0.8.0 and drop it immediately after. In any case I'll make sure the latest including this fix is readily usable. 

I haven't put much thought if the grounding sql queries from the new compiler will perform well in pgxl. At least there's no update anymore. Curious to hear any success or failure story there.

PR #447, opened 2016-01-05
Enhances deepdive load to handle data with subset of columns

netj:
Now it's possible to load a subset of columns like this: `deepdive load "foo(x,y,z)" foo_data.tsv`

Relations declared as random variables in the model are recognized, and when no columns are explicitly specified, the internal label column will be expected to come as the last one but not the internal id column.

Also now groups the data sources by format and calls the driver's `db-load` to make it possible to support parallel loading from each driver.


database/deepdive-load
```diff
@@ -1,14 +1,24 @@
...
-# When SOURCE is unspecified, the data is loaded from files found on path
-# input/RELATION.* under the DeepDive application.
+# > deepdive load RELATION[(COLUMN[,COLUMN]...)] [SOURCE...]
+# Initializes given RELATION in the database and loads data from SOURCE for
+# optionally specified COLUMNs.  When SOURCE is unspecified, the data is loaded
+# from files found on path input/RELATION.* under the DeepDive application.
+#
+# When RELATION is a random variable and no COLUMN is specified, the SOURCE are
+# expected to contain the `label` column at the end, in addition to all
+# user-defined ones.
```
feiranwang: This requires the users to be aware of the `label` column in variable relation, but not declaring it in the schema declaration. IMO it would be better to be more consistent: either the `label` column is explicitly exposed to the users, or it's completely hidden from the users.

netj:
You know I prefer completely hiding them. However do you think it's okay to always require a supervision rule that fills the labels from a separate data source? This may be in fact a good thing that discourages direct supervision with no practical difference. The only things that appear more complicated are the toy examples. If you agree I'm happy to revert the last commit that introduced this inconsistency.

feiranwang:
I think it's ok to keep the current semantics, and it's good to be able to directly load data into variable relations.

PR #449, opened 2016-01-05
Drops Scala codebase and moves things around into better shape

netj:
I'm still deciding whether to keep the Scala codebase around in the next 0.8.0 release, or to drop it and completely move on to this clean state now while pushing the last Scala code in a parallel release 0.7.2.  Keeping the legacy code with backward compatibility will be less surprising and may be easier for the users to gradually adapt the new operations, but it can also delay the switch as it's not mandatory.

Please voice in your thoughts! @alldefector @raphaelhoffmann @Colossus @zifeishan @xiaoling @ThomasPalomares @ajratner


alldefector:
My vote would be to drop scala in 0.8.0 because doing it in 0.8.1 may come as a bigger surprise since it's a minor release (or we'd have to unnecessarily wait till 0.9.0). Cannot wait to see Akka die!

feiranwang:
It would be great to have a changelog that highlights the major differences from the users' perspective. Some major changes for me as a user: user operations (deepdive compile, etc.), extractor output behavior (replace instead of insert), deprecated features (json extractor, etc.).

netj:
@alldefector Your point makes a lot of sense that it'll be a bigger surprise later.  I'd love to drop it asap to move forward.  Anyone else has concern/objection here?

@feiranwang Of course. There'll be a nice changelog that summarizes all the user facing changes, but in a subsequent PR or commit for preparing the release.  In fact, there are so many inconsistencies to docs introduced by this drop.  Many doc pages need rewrite before the release.

feiranwang:
I see. Makes sense!

PR #450, opened 2016-01-05
Compresses factor graph binary files with bzip2

netj:
when grounding, and decompresses on the fly with bzcat when running the
sampler.

This cuts a 2MB (2300279 bytes) factor graph down to 184kB (184224 bytes)
(one produced by the test with spouse_example/ddlog app) with negligible
impact on runtime (or even faster!).

```bash
groundingTime() {
    local log=$1
    tstart=$((sed -n '\@process/grounding/.*/dump@{ p; q; }' | awk '{print $2}') <$log)
    tend=$((sed -n '\@LEARNING EPOCH 0@{ p; q; }' | awk '{print $2}') <$log)
    echo $(date --date="$tend" +%s.%N) - $(date --date="$tstart" +%s.%N) | bc
}

$ # dump from database and load by sampler without compression
$ groundingTime test/postgresql/spouse_example/ddlog/run/20151221/040549.120506000/log.txt
3.363981000

$ # dump from database and load by sampler with compression
$ groundingTime test/postgresql/spouse_example/ddlog/run/20151221/034709.985065000/log.txt
3.319830000
```

Note that gzip has a 4GB limitation, so bzip2 was used despite its
higher computational cost.  xz may be another good candidate to consider.

### DIFF
--- a/compiler/compile-config/compile-config-2.01-grounding
+++ b/compiler/compile-config/compile-config-2.01-grounding
@@ -175,7 +175,7 @@ include "sql";
         varPath=\"$DEEPDIVE_GROUNDING_DIR\"/variable/\(.variableName | @sh)
         mkdir -p \"$varPath\"
         cd \"$varPath\"
-        find . -name 'variables-*.bin' -exec rm -rf {} +
+        find . -name 'variables.part-*.bin.bz2' -exec rm -rf {} +
         export DEEPDIVE_LOAD_FORMAT=tsv
 
         # dump the variables, joining the holdout query to determine the type of each variable
@@ -248,7 +248,7 @@ include "sql";
                 ]
             } | asSql | @sh) \\
             command=\("
-                format_converter variable /dev/stdin >(bzip2 >variables.part-${DEEPDIVE_CURRENT_PROCESS_INDEX}.bin.bz2)
+                format_converter variable /dev/stdin >(pbzip2 >variables.part-${DEEPDIVE_CURRENT_PROCESS_INDEX}.bin.bz2)
             " | @sh) \\
             output_relation=
         "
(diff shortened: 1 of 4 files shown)


feiranwang:
Maybe we should test it on a larger factor graph (> 1GB) to see how it performs?

netj:
Compression certainly has overhead.  The question is whether it'll be a bottleneck.  I'm trying to ground a larger one by running spouse example on a larger corpus I synthesized.  But it revealed mkmimo's lower throughput, and running much slower than expected.

Meanwhile, here're my notes from doing a quick overhead test with several choices: https://gist.github.com/netj/c6f15bb78ff3a52057cb

netj:
Before I forget, I'll drop some numbers I got a while ago for a large factor graph I synthesized by duplicating the corpus for the spouse example (~12GB uncompressed, 199k vars, 16k weights, 337M factors).
```
LOADED VARIABLES: #199907
         N_QUERY: #139603
         N_EVID : #60304
LOADED WEIGHTS: #16664
LOADED FACTORS: #337742718
```

Following are rough measurements on raiders6 with 111 processes, only accounting the dumping time and loading time.

### uncompressed
* 11828322038 bytes (~12GiB)
* 401.224535 secs

### pbzip2
* 197572897 bytes (~191MiB; 59.8x smaller)
* 420.276131 secs (+19s; +4.7% increase)

### bzip2
* 195875810 bytes (~189MiB; 60.4x smaller)
* 464.805231 secs (+64s; +16% increase)

Since the full grounding took significantly more time (materializing the factors, weights), I'd say compression overhead is negligible while it's savings on storage footprint and in turn I/O are quite dramatic.  The higher-than-usual compression rate (>>10x) is probably due to the regularity in the factor graph data representation.  I think we should turn this on by default unless there's a really good counter argument.

feiranwang:
Seems there's a huge saving in space with negligible overhead! Merging.

PR #451, opened 2016-01-06
Fixed weight table for Postgres-XL

zifeishan:
PGXL doesn't support correlated UPDATE involving multiple nodes:
http://sourceforge.net/p/postgres-xc/mailman/message/26727433/

This workaround prevents the issue by materializing the table first. It should work for both PGXL and other databases.
See #446 for more discussions.

### DIFF
--- a/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
+++ b/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala
@@ -913,13 +913,13 @@ trait SQLInferenceRunner extends InferenceRunner with Logging {
       DROP TABLE IF EXISTS __dd_tmp_${WeightsTable};
       CREATE TABLE __dd_tmp_${WeightsTable} AS
       SELECT w.id, fw.weight
-      FROM ${WeightsTable} w, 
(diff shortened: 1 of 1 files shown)


alldefector:
@netj should this go to 0.8?

netj:
I thought we were dropping Scala, so this literal PR won't go into 0.8, but into 0.7.2 or so.

Reusing learned weights is one of the two missing features in the new compiler (the other being GP's parallel unloading/loading).  I'm planning to implement them pretty soon but other priorities are starting to invade my calendar so it may take more than a week.  However, I can make sure this kind of UPDATE does not show up in the compiled grounding code by doing a simple OUTER JOIN when dumping the weights.

alldefector:
Got it. Thanks!

zifeishan:
@netj Regarding 0.8 vs 0.7.2, will 0.7.2 include the grounding improvements that's currently in master branch? That is valuable since it doubles the grounding performance in our setting.

netj:
@zifeishan Sure, I'll backport most changes relevant to Scala code to 0.7.x and release 0.7.2.  So let's get this in to master, then freeze the Scala codebase in master.

PR #452, opened 2016-01-08
fix db-create for greenplum

ThomasPalomares:
While creating a table during a deepdive run, the driver db-create currently create a syntax containing "include all", which is not supported by Greenplum. 

This PR simply modifies the alias of greenplum/db-create from postgresql/db-create to mysql/db-create


netj:
LGTM Thanks for this fix!

PR #453, opened 2016-01-12
Refactors `deepdive create`

netj:
Refactors `db-create` into several pieces that are easier to extend/special-case by each db-driver:
* `db-create-table-def`
* `db-create-table-like`
* `db-create-table-as`
* `db-create-view-as`

And turns it into a user-facing command `deepdive-create` which also takes care of creating relations defined in app.ddlog or schema.json.

Greenplum driver now preserves the `DISTRIBUTED BY` columns when creating a new table `LIKE` another.

Renames `init/db` to `init/app` and deprecates `deepdive-initdb` (in favor of `deepdive db init` and `deepdive create`).


ThomasPalomares:
@chrismre: this fixes the issue about distributed_by in temporary tables. 
With that + a bug in mkimo that Jaeho noticed, the time consuming extractions are now ~50x faster

chrismre:
AWESOME! :)

PR #456, opened 2016-01-15
Fixes deepdive load to handle large .sql files

ThomasPalomares:
Fixes deepdive load to handle large .sql files


chrismre:
cc: @HazyResearch/genomics Does this fix the issue from today's meeting?

Colossus:
No

ThomasPalomares:
No, the problem mentioned today is fixed in https://github.com/HazyResearch/deepdive/pull/453. This PR is for another (smaller issue)

PR #457, opened 2016-01-15
Updates mkmimo that doesn't sacrifice throughput by default

netj:
with the downside of having busy waiting on slower UDFs


netj:
@chrismre @HazyResearch/genomics This is the fix for the throughput issue @ThomasPalomares explained in #453

PR #459, opened 2016-01-19
Fixes deepdive-load test with huge.sql to be lighter

netj:


ThomasPalomares:
Looks great @netj ! 
make test ONLY=test/postgresql/deepdive_load.bats runs without any failure and it loads correctly the sentences_input.sql file in the genomics application, is it enough to merge this PR or is there another test I should try first ?

netj:
Nope, Travis also confirms everything's fine.  The test failed fast when prior to our fix, but since the fix it was INSERTing too much rows unnecessarily and made the test too costly so I tweaked it a bit.  Please merge if it LGTY

PR #472, opened 2016-01-30
Fixes bug in grounding that could crash sampler

netj:
after changing degree of parallelism (DEEPDIVE_NUM_PROCESSES)


ajratner:
Thanks for this :)!

PR #473, opened 2016-01-30
Parallel loading unloading support

netj:
#### Parallel loading
* `deepdive-load` now takes multiple source files and loads in parallel by running multiple `db-load` (i.e., `deepdive db load`) processes which in turn creates multiple connections to the database server.
    * postgresql driver uses multiple connections, and greenplum, pgxl shares the implementation.
    * greenplum could use `gpload` by default but we can extend this later.

#### Parallel unloading
`db-unload` (or equivalently `deepdive db unload`) takes arbitrary SQL query and zero or more sinks/destination files to unload data in parallel.
* postgresql driver first partitions the result set, then uses multiple connections to unload each partition at the same time.
* To avoid all connections performing the same join, it tries to look for commas and `JOIN`s after `FROM` in the query to materialize the result set first.
* This materialization can be controlled by setting `DEEPDIVE_UNLOAD_MATERIALIZED` environment to `true` or `false`.
* greenplum driver uses gpfdist with its `CREATE EXTERNAL TABLE`.
* However, gpfdist's throughput is very low when compared to using postgresql driver's parallel connection approach (1-2MB/s vs. >50MB/s).  The reason we can't use postgresql's naive approach for greenplum or pgxl (by setting `GPFDIST_DISABLE=true`) is because the `LIMIT/OFFSET` does not give the correct partition unless there is a stable total ordering in the result set which is still hard to get with a `ORDER BY`.  In fact, postgresql can also give incorrect result (missing rows, duplicate rows) unless the tables stay intact while unloading in parallel.  I have some ideas to try (written as comments), but first let's put in something and improve later.

Newly added `deepdive-unload` command makes unloading handy for relations declared in the schema into files.
* Unload format is recognized by extension, e.g., `deepdive unload sentences a.tsv b.tsv.bz2 c.csv` or `DEEPDIVE_LOAD_FORMAT` environment.
* When more than one sinks/destination files are given that share the same format, unloads in parallel.

#### Not enabled by default for running extractors yet
* `DEEPDIVE_NUM_PARALLEL_LOADS` and `DEEPDIVE_NUM_PARALLEL_UNLOADS` environment variables control the degree of parallelism for loading/unloading while running UDFs through local compute driver (or `deepdive compute`).
* Currently, both are kept to 1 by default to not cause any trouble but we can increase this later to a reasonable number (like 2 maybe?).


util/partition_integers
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
+# partition_integers -- Partitions integer range into even ranges
+##
+import sys
+
+M = int(sys.argv[1])
+N = int(sys.argv[2])
+
+k = float(M)/N
```
SenWu: Can we add a check for N here? E.g. N=0.
Also, if M=10 and N=20, the answer is
1 0 0
0 1 0
1 1 1
0 2 1
1 2 2
0 3 2
1 3 3
0 4 3
1 4 4
0 5 4
Is this answer correct?

SenWu:
Only one comment for corner cases. Looks good other than this.

util/partition_integers
netj: Thanks for catching this.  I wasn't careful enough to include unit tests for this.  I'll make sure it produces error on such case.  Yes if we want more partitions than the number of elements, it'll end up with something like that.  I think it'll produce the a correct partition that covers all elements, right?

SenWu: Yes, it will produce the correct answer. Could it better to set N as M is N>M?

On Feb 1, 2016, at 12:22 AM, Jaeho Shin <notifications@github.com<mailto:notifications@github.com>> wrote:


In util/partition_integers<https://github.com/HazyResearch/deepdive/pull/473#discussion_r51389627>:

> @@ -0,0 +1,16 @@
> +#!/usr/bin/env python
> +# partition_integers -- Partitions integer range into even ranges
> +##
> +import sys
> +
> +M = int(sys.argv[1])
> +N = int(sys.argv[2])
> +
> +k = float(M)/N


Thanks for catching this. I wasn't careful enough to include unit tests for this. I'll make sure it produces error on such case. Yes if we want more partitions than the number of elements, it'll end up with something like that. I think it'll produce the a correct partition that covers all elements, right?

SenWu:
LGTM.

netj:
Thanks for merging but it was failing at Travis.  Not sure why it's failing there, but in the future, let's only merge when everything's green.

PR #468, opened 2016-01-25
Various usability improvements

netj:
Corrects many glitches and annoyances found so far while working with early adopters. (Thanks everyone who suffered!)

* Better bash completion for deepdive commands with easy instructions.
* `deepdive run` and `deepdive initdb` have been added back as shorthand and backward compatibility.
* `deepdive redo` won't lose last state if canceled.
* Compiler won't complain about DDlog rules for VIEWs.
* `process/init/app` always precedes all other processes.
* No more silly oscillating progress bar, only numbers are shown.
* Support for environment variables in db.url.
Please see the individual commit messages for more.

Issue #416, opened 2015-09-27
We should output "supervised" data points as well


alldefector:
Looks like data points with filled is_correct values get excluded from Gibbs sampling. But the expectations on those data points are useful for both dev process and the final data product.

Right now, people would have to cumbersomely make copies of those supervised data points to force DD to run inference on them (@ajratner @zhangce @raphaelhoffmann for example). I think I've brought this up a while ago.

Do you agree that we should always include supervised data points in inference? I realize that we may also need to change code of the sampler for this. Would that be a lot of work?

zhangce:
@thodrek 's new framework should provide an ultimate cleaner solution here. For short term, it should also be an easy fix.

zifeishan:
Agreed that this is very useful in many applications.

On Sun, Sep 27, 2015 at 2:33 AM, zhangce <notifications@github.com> wrote:

> @thodrek <https://github.com/thodrek> 's new framework should provide an
> ultimate cleaner solution here. For short term, it should also be an easy
> fix.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/HazyResearch/deepdive/issues/416#issuecomment-143507063>
> .
>



-- 
Zifei Shan
M.S. in Computer Science, Stanford University (2015)

raphaelhoffmann:
@zhangce @SenWu @feiranwang Can you show us a few lines of ddlog code that does this copying of supervised examples? Or should we already do it inside our UDFs?

feiranwang:
@raphaelhoffmann This can be achieved by turning on the sampler argument `--sample_evidence`, which will let the sampler output probabilities for evidence variables as well. The sampler argument could be added in `deepdive.conf`, and when the `app.ddlog` is compiled by `deepdive run`, this option will get incorporated.

zhangce:
@feiranwang Thanks, Feiran! Yeah, I forget you added this before...

ajratner:
Yeah sorry for not chiming in earlier, we ran into this problem in genomics and @feiranwang added the `--sample_evidence` flag for us pre-ddlog, so we had been using it

Confirming that there is a simple way to incorporate this with ddlog would be great too!

raphaelhoffmann:
Thanks! That works! @alldefector I think we can close this issue.

alldefector:
Hi @feiranwang , did you mean we can specify `--sample_evidence` in `app.ddlog` or the `deepdive run` command line? I wasn't able to find out how...

[...]

PR #475, opened 2016-01-30
New spouse example

### LINKED ISSUE #236: Encoding / Decoding PostgreSQL Arrays for tsv_extractor
Currently if users want to use arrays as input of TSV extractor, they typically use "array_to_string" to add delimiters. It is also not easy for users to figure out how to dump arrays in outputs of TSV extractors.

We should provide ways in a library (ddlib?) to encode / decode PostgreSQL arrays for TSV extractors. There are some code in some of our applications.

netj:
This is the new spouse example we wrote from scratch.  Here's the [hackathon notes](https://docs.google.com/document/d/1gPwToancyVxUbQaM1erxaip-JY9ADuPKIwypJKYwCN4/edit?usp=sharing) fyi.
- starting from raw corpus (.jsonl)
- including NLP markup process
- distant supervision using DDlog as well as UDF (data programming!)
- includes a much cleaner way to write UDFs
- mindtagger and mindbender bits

@feiranwang The README is still incomplete and the example code needs a bit more scrubbing, but we can do this in the upcoming doc-athon.  Although @ajratner and @thodrek and I did this together, I wanted an independent person to have another look.  This PR actually depends on #468, so you may want to wait until it's merged, or just compare against that branch to see the diff more clearly.  Chunking example seems to be failing in the tests, and I'll look into that. (No idea why a seemingly independent piece suddenly broke.)

Resolves #236.


compiler/deepdive-default.conf
```diff
@@ -12,7 +12,7 @@ deepdive {
   }
 
   sampler.sampler_cmd: "sampler-dw"
-  sampler.sampler_args: "-l 1000 -s 1 -i 1000 --alpha 0.01"
+  sampler.sampler_args: "-l 1000 -s 1 -i 1000 --alpha 0.01 --sample_evidence"
```
feiranwang: Seems some different voices there #416. So we're decided to set it as default?

examples/spouse/app.ddlog
```diff
@@ -0,0 +1,243 @@
...
+
+# Features
+@weight(f)
+has_spouse(p1_id, p2_id) :-
+  spouse_candidate(p1_id, _, p2_id, _),
+  spouse_feature(p1_id, p2_id, f).
+
+# Inference rule: Symmetry
+@weight(3.0)
+has_spouse(p1_id, p2_id) => has_spouse(p2_id, p1_id) :-
```
feiranwang: Should this be equal?

examples/spouse/deepdive.conf
```diff
@@ -0,0 +1,2 @@
+deepdive.calibration.holdout_fraction:0.6
```
feiranwang: 60% for calibration is a bit uncommon. Just to double check, is this intended?

examples/spouse/input/articles.tsv.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$0")"
+
+cat /dfs/scratch0/thodrek/signalmedia/signalmedia-1m.jsonl |
```
feiranwang: Hardcoded path here.

compiler/deepdive-default.conf
thodrek: We are using it here because of the noisy rules. To be honest I don't think we should have it on by default but that's just my opinion.

examples/spouse/deepdive.conf
thodrek: We started by testing a very small corpus and had that so that we could have a sufficient number of "spouse" instances in the test data. This was without using --sample_evidence.

compiler/deepdive-default.conf
netj: @feiranwang Good catch!  @thodrek This is actually the default settings we updated, not just our example.

We've seen too many people bitten by this default setting, and enough is enough.  Whoever that's rigorous enough wanting to exactly evaluate the system's performance won't be misled by the extra inference results.  So I insist on turning this option on by default.  However, I'm all for adding a clear way to prevent misinterpretation, so let's discuss what can be the best additional output, e.g., what extra view can we create and/or should we compute two numbers that excludes/includes holdout?

examples/spouse/app.ddlog
netj: I agree.  I think the only reason implication appeared here is because the other example we looked at was written before we added the Equals-factor syntax.

examples/spouse/input/articles.tsv.sh
netj: Right.  We should change this to grab from the same dir (`input/`) and put instructions to download the full corpus (people have to go through license agreement and get an email with download link btw).

We'll grab a subset of articles, get NLP markups and include the sentences.tsv.bz2 instead for running our integration tests.

netj:
@ajratner This now has all green lights.  You can merge this whenever you want and I'll merge master into docathon-v0.8 so everyone's on the latest code.

PR #479, opened 2016-02-05
Fixes `deepdive compute` to `deepdive compute execute`

netj:
to leave room for the sub-command for future extensions by compute drivers

### DIFF
--- a/compiler/compile-code/compile-code-tsv_extractor
+++ b/compiler/compile-code/compile-code-tsv_extractor
@@ -20,7 +20,7 @@ cd \"$(dirname \"$0\")\"
 export DEEPDIVE_CURRENT_PROCESS_NAME=\(.name | @sh)
 export DEEPDIVE_LOAD_FORMAT=tsv
 
-deepdive compute \\
+deepdive compute execute \\
     input_sql=\(.input | @sh) \\
     command=\(.udf | @sh) \\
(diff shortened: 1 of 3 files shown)


avantgarde1310:
@netj Should I do anything else on this? I need to pull this to my branch eventually, right?

netj:
Please take a look, comment, then merge once everything looks good to you. Standard code review :)

You can then perhaps rebase your branch.

avantgarde1310:
Looks good to me! Just some questions below. 

When we were discussing the idea of having other compute-* (such as compute-status, compute-loadin, compute-unloadout, compute-submit) commands, we also talked about requiring DeepDive be installed on the submission nodes as well. I was wondering if running `deepdive compute status` (for example) should be allowed on both our workstation (laptop, etc.) and the submission node? 

Or would it only make sense to run these compute-* commands inside the context of compute-execute (i.e. compute-* commands would be "private functions" being used only by compute-execute)?

netj:
It'll largely be up to the driver but that'd be nice feature to have. `compute-status` for example can show all running processes for the config, right? I'd recommend hiding private commands under a clear namespace such as `compute-remote-` and let others be end user friendly by default.

netj:
Rebased, fixing a few merge conflicts.

igozali:
LGTM

PR #481, opened 2016-02-09
`deepdive-model` for reusing weights and expediting other model tasks

### LINKED ISSUE #463: categorical variables do not work with skip_learning
When using `skip_learning` and loading a weight_table, the weights are assigned by joining only on the  `description` column but not the `cardinality` column. See the code [here](https://github.com/HazyResearch/deepdive/blob/e770f7e9ea7ff71dedc05e6392349f459807bbfd/src/main/scala/org/deepdive/inference/SQLInferenceRunner.scala#L913).

Consequently, the same feature weights are assigned to every category, eg.:
```
  3958 |       0 | -0.0466067 | 00003       | factor_feature-^_^
  3957 |       0 | -0.0466067 | 00002       | factor_feature-^_^
  3956 |       0 | -0.0466067 | 00001       | factor_feature-^_^
  3955 |       0 | -0.0466067 | 00000       | factor_feature-^_^
  3959 |       0 | -0.0466067 | 00004       | factor_feature-^_^
 37944 |       0 |   -1.91256 | 00004       | factor_feature-¨
 37940 |       0 |   -1.91256 | 00000       | factor_feature-¨
 37943 |       0 |   -1.91256 | 00003       | factor_feature-¨
 37942 |       0 |   -1.91256 | 00002       | factor_feature-¨
 37941 |       0 |   -1.91256 | 00001       | factor_feature-¨
``` 
This is wrong; the fix is to consider both cardinality and description when loading weights.

@zifeishan and I are fixing this in our branch for now. Does this issue also exist in DD0.8?

netj:
Shorthand commands for repeating a group of processes related to inference.
* `deepdive model ground`
* `deepdive model learn`
* `deepdive model infer`

Commands for keeping and reusing weights.
* `deepdive model weights keep`
* `deepdive model weights reuse`
* `deepdive model weights list`
* `deepdive model weights drop`
* `deepdive model weights init`

(These work with factor graphs as well if you replace `weights` with `factorgraph` but more bits should be kept to make it fully useful/meaningful, e.g., speeding up grounding, etc.  Currently it just stashes/restores the dumped factorgraph binaries but that's quite useless unless we keep around the database tables for grounding as well.)

Here's a hypothetical workflow:
```bash
deepdive do init/app

# learn weights from a small corpus and keep it for future reuse
deepdive create table articles
deepdive load articles /data/small/corpus.tsv.bz2
deepdive mark new articles
deepdive model learn  # taking care of all necessary extraction and grounding
deepdive model weights keep foo

# load larger corpus
deepdive create table articles
deepdive load articles /data/LARGE/corpus.tsv.bz2
deepdive mark new articles

# reuse the weights (skipping learning) and do inference
deepdive model weights reuse foo  # taking care of all extraction and grounding for larger data
deepdive model infer
```

I'll soon add a test that looks like above for a few of our examples (either in this PR or another).

Closes #463

### DIFF
--- a/inference/deepdive-model
+++ b/inference/deepdive-model
@@ -74,16 +74,24 @@ case $Command in
                         rsync -avH --delete --copy-unsafe-links run/model/factorgraph{,.done} "$dest"/
                         ;;
                     weights)
+                        # automatically load weights only if learning is already done
+                        if ! deepdive-"done" data/model/weights &&
+                             deepdive-"done" process/model/learning; then
+                            DEEPDIVE_PLAN_EDIT=false \
+                            deepdive-"do" data/model/weights
+                        fi
+                        # dump non-zero weights
                         show_progress output_from "$dest" -- \
-                        deepdive sql eval "
-                            SELECT id
-                                 , isfixed
-                                 , initvalue
-                                 , description
-                                 , categories
-                              FROM dd_graph_weights
-                             WHERE initvalue <> 0
-                        " format=csv | pbzip2 >"$dest"/dd_graph_weights.csv.bz2
+                        deepdive-sql eval "
+                            SELECT description
+                                 , ARRAY_AGG(weight)
+                                 , ARRAY_AGG(categories::text)
+                              FROM dd_inference_result_weights wo
+                                 , dd_graph_weights            wi
+                             WHERE wo.id = wi.id
+                               AND weight <> 0
+                             GROUP BY description
+                        " format=tsv | pbzip2 >"$dest"/dd_graph_weights.tsv.bz2
                         ;;
                 esac
                 # keep a LAST symlink pointing to the last
--- a/inference/deepdive-model
+++ b/inference/deepdive-model
@@ -30,7 +30,7 @@ case $Command in
         ;;
 
     learn)
-        exec deepdive-redo process/model/learning data/model/weights
+        exec deepdive-redo process/model/learning data/model/{weights,probabilities}
         ;;
 
     infer)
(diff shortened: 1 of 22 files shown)


inference/deepdive-model
```diff
@@ -0,0 +1,242 @@
...
+                                                SELECT description
+                                                     , UNNEST(categories) AS categories
+                                                     , UNNEST(weights)    AS weight
+                                                  FROM ${weightTableName}_array
+                                            "
+                                            deepdive sql "DROP TABLE ${weightTableName}_array"
+                                            ;;
+                                        *.tsv.bz2)
+                                            # TODO parallel load
+                                            deepdive load "$weightTableName" "$weightFile"
```
raphaelhoffmann: Does this work? I ran `deepdive model weights reuse foo` on the spouse tsv example, and it I got
```
ERROR:  missing data for column "weight"
CONTEXT:  COPY dd_graph_weights_reuse, line 1: "f_has_spouse_features--word_between=D	0.723347"
```
Table `dd_graph_weights_reuse` has three columns, but the dumped file that we are trying to load only has two columns.

netj: Sorry, I may have left some bugs after last minute changes for multinomial.  Another reason to set up the tests first.  It should probably work if replaced with the following line.
```
deepdive load "$weightTableName(description,weight)" "$weightFile"
```

inference/deepdive-model
```diff
@@ -0,0 +1,240 @@
...
+                                    [[ -e "$weightFile" ]] || continue
+                                    case $weightFile in
+                                        *.multinomial.tsv.bz2)
+                                            # first load the grouped data from file
+                                            deepdive create table "${weightTableName}_array" "description:TEXT" "categories::TEXT[]" "weights:DOUBLE PRECISION[]"
+                                            # TODO parallel load
+                                            deepdive load "${weightTableName}_array" "$weightFile"
+                                            # explode the group
+                                            # XXX UNNEST is postgres-specific
+                                            deepdive sql "INSERT INTO $weightTableName
```
raphaelhoffmann: it probably doesn't matter in practice but it might be more efficient to do `create table as`

raphaelhoffmann:
This works really great! Ready to merge.

inference/deepdive-model
netj: Since we had more than one sources of weights (one for boolean, another for multinomial factors) I kept it as INSERT, but we could also create two tables and add a UNION view if you think there's a big difference.

netj:
@raphaelhoffmann Thanks for testing and reviewing it!  I'll press the button myself.

PR #496, opened 2016-02-21
Fixes bug in fast_seqassign for Greenplum

### LINKED ISSUE #495: Greenplum DB driver ID assignment is buggy for categorical variables
We ran the chunking example with DD 0.8 on Greenplum (with 20 segments) but always get this error:
```
2016-02-21 00:24:11.464530 ################################################
2016-02-21 00:24:11.661727 LOADED VARIABLES: #270052
2016-02-21 00:24:11.661813          N_QUERY: #49389
2016-02-21 00:24:11.661828          N_EVID : #220663
2016-02-21 00:24:11.722552 LOADED WEIGHTS: #268970
2016-02-21 00:24:12.756042 LOADED FACTORS: #1049594
2016-02-21 00:24:12.995216 sampler-dw.bin: src/dstruct/factor_graph/factor_graph.cpp:375: void dd::FactorGraph::safety_check(): Assertion `this->weights[i].id == i' failed.
2016-02-21 00:24:13.171646 process/model/learning/run.sh: line 22: 184415 Aborted                 (core dumped) 
```

The corresponding code verifies that the weight vector is loaded in the order of id = 0, 1, 2, ...:
https://github.com/HazyResearch/sampler/blob/master/src/dstruct/factor_graph/factor_graph.cpp#L395

The `dump_weights/run.sh` scripts run queries similar to
```
select * from dd_weightsmulti_inf_istrue_tag limit 10;
```

While the above query returns results consistently in the same order for PG, the order can vary if we run it on GP again and again. So the weight vector ordering seems to have relied on an implementation detail of Postgres, and now it's broken in Greenplum (we are running latest GP release as of today).

Interestingly, we could run the spouse example on GP. With spouse, the first column is `isfixed` -- which seems to be always `f`. With chunking, the first column of `dd_weightsmulti_inf_istrue_tag` is `dd_weight_column_0`, which could take on different values. And GP by default distributes on the first column. So that seems to explain the ordering issue:
```
# select * from dd_weightsmulti_inf_istrue_tag limit 3;
 dd_weight_column_0 | isfixed | initvalue | id  | categories 
--------------------+---------+-----------+-----+------------
 word=Results       | f       |         0 | 296 | 0
 word=obliged       | f       |         0 | 307 | 0
 word=Eurocom       | f       |         0 | 318 | 0
(3 rows)

# select * from dd_weightsmulti_inf_istrue_tag limit 3;
 dd_weight_column_0 | isfixed | initvalue | id  | categories 
--------------------+---------+-----------+-----+------------
 word=Leahy         | f       |         0 | 170 | 0
 word=banana        | f       |         0 | 176 | 0
 word=9.875         | f       |         0 | 182 | 0
(3 rows)

# select * from dd_weightsmulti_inf_istrue_tag limit 3;
 dd_weight_column_0 | isfixed | initvalue | id  | categories 
--------------------+---------+-----------+-----+------------
 word=Results       | f       |         0 | 296 | 0
 word=obliged       | f       |         0 | 307 | 0
 word=Eurocom       | f       |         0 | 318 | 0
(3 rows)
```

Between adding `ORDER BY` to the dump_weights code and revising the sampler, which would be a better solution?

netj:
Fixes #495 as well as related bugs and adds tests.

Does anyone have some numbers comparing SEQUENCE vs. this Python UDF? @zhangce @feiranwang @SenWu  Is GP SEQUENCE that slow?  I remember in PGXL's case, SEQUENCE was completely broken (all data nodes were assigning duplicates), but if GP managed to support it, I think it's better to rely on it.

### DIFF
--- a/database/db-driver/greenplum/db-assign_sequential_id
+++ b/database/db-driver/greenplum/db-assign_sequential_id
@@ -17,73 +17,66 @@ Table=$1 Column=$2 BeginId=$3 Increment=${4:-1}
 db-execute "CREATE LANGUAGE plpgsql;" || true
 db-execute "CREATE LANGUAGE plpythonu;" || true
 db-execute "
-    CREATE OR REPLACE FUNCTION clear_count_1(sid INT) RETURNS INT AS
-    \$\$
-    if '__count_1' in SD:
-      SD['__count_1'] = -1
-      return 1
-    return 0 # XXX why return a confusing 0 when it still counts toward COUNT()?
+    CREATE OR REPLACE FUNCTION fast_seqassign_init()
+    RETURNS INT AS \$\$
+      if 'cumulative_cnt' in SD:
+        SD.pop('cumulative_cnt')
+      return 0
     \$\$ LANGUAGE plpythonu;
 
-    CREATE OR REPLACE FUNCTION updateid(startid BIGINT, increment BIGINT, sid INT, sids INT[], base_ids BIGINT[], base_ids_noagg BIGINT[]) RETURNS BIGINT AS
-    \$\$
-    if '__count_1' in SD:
-      a = SD['__count_2']
-      b = SD['__count_1']
-      SD['__count_2'] = SD['__count_2'] - 1
-      if SD['__count_2'] < 0:
-        SD.pop('__count_1')
-
-    else:
-      for i in range(0, len(sids)):
-        if sids[i] == sid:
-          SD['__count_1'] = base_ids[i] - 1
-          SD['__count_2'] = base_ids_noagg[i] - 1
-      a = SD['__count_2']
-      b = SD['__count_1']
-      SD['__count_2'] = SD['__count_2'] - 1
-      if SD['__count_2'] < 0:
-        SD.pop('__count_1')
-
-    return startid + increment*(b-a)
-
+    CREATE OR REPLACE FUNCTION fast_seqassign_next(startid BIGINT, increment BIGINT, this_gpsid INT, gpsids INT[], cumulative_cnts BIGINT[], cnts BIGINT[])
+    RETURNS BIGINT AS \$\$
+      if 'cumulative_cnt' not in SD:
+        for gpsid, cumulative_cnt, cnt in zip(gpsids, cumulative_cnts, cnts):
+          if gpsid == this_gpsid:
+            # assignment is done using the cumulative count for this segment and a decrementing counter from the size of this segment down to one
+            SD['cumulative_cnt'] = cumulative_cnt
+            SD['remaining'] = cnt
+            break
+      if 'cumulative_cnt' in SD and SD['remaining'] > 0:
+        id = startid + increment * (SD['cumulative_cnt'] - SD['remaining'])
+        SD['remaining'] = SD['remaining'] - 1
+        if SD['remaining'] <= 0:
+          SD.pop('cumulative_cnt')
+        return id
+      else:
+        # XXX no segment was found
+        raise plpy.ERROR('No non-empty id range was assigned for segment %s in the initial partition' % str(this_gpsid))
     \$\$ LANGUAGE plpythonu;
 
-    CREATE OR REPLACE FUNCTION fast_seqassign(tname character varying, cname character varying, startid bigint, increment bigint) RETURNS TEXT AS
-    \$\$
+    CREATE OR REPLACE FUNCTION fast_seqassign(tname CHARACTER VARYING, cname CHARACTER VARYING, startid BIGINT, increment BIGINT)
+    RETURNS TEXT AS \$\$
     BEGIN
-      EXECUTE 'DROP TABLE IF EXISTS tmp_gpsid_count       CASCADE;';
-      EXECUTE 'DROP TABLE IF EXISTS tmp_gpsid_count_noagg CASCADE;';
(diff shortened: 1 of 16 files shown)


alldefector:
We also need to add the `inc` parameter (as in PG's version this script).

netj:
@alldefector PTAL

database/db-driver/greenplum/db-assign_sequential_id
```diff
@@ -1,97 +1,93 @@
...
+          if gpsid == this_gpsid:
+            # assignment is done using the cumulative count for this segment and a decrementing counter from the size of this segment down to one
+            SD['cumulative_cnt'] = cumulative_cnt
+            SD['remaining'] = cnt
+            break
+      if 'cumulative_cnt' in SD and SD['remaining'] > 0:
+        id = startid + increment * (SD['cumulative_cnt'] - SD['remaining'])
+        SD['remaining'] = SD['remaining'] - 1
+        if SD['remaining'] <= 0:
+          SD.pop('cumulative_cnt')
```
alldefector: Looks like these two lines would make the next call re-initialize `SD['cumulative_cnt']`, which in turn would circumvent the `else` block from ever being activated. I think we should just remove them.

database/db-driver/greenplum/db-assign_sequential_id
```diff
@@ -1,97 +1,93 @@
...
+    CREATE OR REPLACE FUNCTION fast_seqassign_next(startid BIGINT, increment BIGINT, this_gpsid INT, gpsids INT[], cumulative_cnts BIGINT[], cnts BIGINT[])
+    RETURNS BIGINT AS \$\$
+      if 'cumulative_cnt' not in SD:
+        for gpsid, cumulative_cnt, cnt in zip(gpsids, cumulative_cnts, cnts):
+          if gpsid == this_gpsid:
+            # assignment is done using the cumulative count for this segment and a decrementing counter from the size of this segment down to one
+            SD['cumulative_cnt'] = cumulative_cnt
+            SD['remaining'] = cnt
+            break
+      if 'cumulative_cnt' in SD and SD['remaining'] > 0:
```
alldefector: This may not be consequential, but since this UDF is invoked once per row, maybe `elif SD['remaining'] > 0:`?

database/db-driver/greenplum/db-assign_sequential_id
netj: I'm not sure if I understood you correctly.  The else block below is for making loud noise when an empty partition is assigned or no partition is assigned for the segment.  Maybe instead of popping, we can put an assertion here that remaining should never go negative.

database/db-driver/greenplum/db-assign_sequential_id
netj: Then most of the code here including `id = ...` and `return id` should be duplicated in the first `if`, no?

database/db-driver/greenplum/db-assign_sequential_id
```diff
@@ -1,97 +1,93 @@
...
+"
 
+else # if either plpgsql or plpythonu is not available
 # Fall back to using PostgreSQL sequence generator named after the table and column
 # See: http://www.postgresql.org/docs/current/static/sql-createsequence.html
 seq="dd_seq_${Table}_${Column}"
 deepdive sql "
     DROP SEQUENCE IF EXISTS $seq CASCADE;
-    CREATE TEMPORARY SEQUENCE $seq MINVALUE -1 START $BeginId;
+    CREATE TEMPORARY SEQUENCE $seq INCREMENT BY $Increment MINVALUE $(($BeginId - 1)) START $BeginId;
```
alldefector: What is `- 1` for?

netj: It's a relic from the Scala code I based my rewrite on.  I wasn't sure if other parts were depending on the particular value, but we can certainly take that out since I'm pretty sure it's unnecessary.

database/db-driver/greenplum/db-assign_sequential_id
alldefector: `SD['remaining'] == 0` is the acceptable case. After that, there should be no more calls. However, if you do `SD.pop` here, the next call would enter the above `if 'cumulative_cnt' not in SD:` branch instead of the `raise` statement.

If you just remove these two lines, the next call would go to the `raise` statement.

database/db-driver/greenplum/db-assign_sequential_id
alldefector: I see... Maybe check `'cumulative_cnt' in SD` at the end of the first `if` block then.

For further optimization, we could also get rid of the multiplication below. Something like
```
id = SD['cur_val']
SD['cur_val'] += increment
return id
```

database/db-driver/greenplum/db-assign_sequential_id
alldefector: Yeah, the docs suggest that `minvalue` is usually equal to `start`.
http://www.postgresql.org/docs/8.3/static/sql-createsequence.html

database/db-driver/greenplum/db-assign_sequential_id
netj: Makes sense. We can turn the else block to handle both out-of-range cases if you think that's clearer.

alldefector: Yep :) (Looks like it current does that already...)

database/db-driver/greenplum/db-assign_sequential_id
netj: Sounds good.  I see your concern for checking SD twice on the hot-path.  I'll also get rid of redundant ops.  I guess that was my original intention, but my brain got overheated decrypting the original code... :)

alldefector:
LGTM!

PR #498, opened 2016-02-22
bug fixed in ddlib/util.py for parsing text[]

ThomasPalomares:
Bug when writing an udf with @tsv_extractor that has a "text[]" as an input. The variable "tok" doesn't appear in the rest of the document, certainly a relic from an old code.


netj:
@ThomasPalomares can you expand the test to include the input that triggered the bug and expected output?

But wait.. I was looking at [deepdive_udf_util.bats](https://github.com/HazyResearch/deepdive/blob/fix_bug_parser_in_ddlib/test/postgresql/deepdive_udf_util.bats) and it seems it's just a duplicate of deepdive_sql.bats without the relevant tests!  @ajratner I thought you added some tests somewhere.. or maybe I was hallucinated?

It should be fairly simple to tailor that bats file to test ddlib.util if we just take the `NastyTSV` and feed it to an identity Python UDF that uses `@tsv_extractor` and `@returns`, etc.

ajratner:
Sorry, I think I started but got interrupted by something else... @ThomasPalomares lmk if I can help here, will be around lab today

ThomasPalomares:
Hi @netj and @ajratner, 
thanks for the comments. I tried to create a test for parsing the NastyTSV but I have two issues:
 - the parsing fails when there is a 'boolean' in @tsv_extractor or @returns (surprisingly since it is only 't' and 'f' in the nasty TSV).
 - it works when I modify the booleans to the type 'text', but then fails on the last column of the NastyTSV. I am not sure why (and I am a bit lost in this last array with all the "/" and escapes), let me know if you know why it fails !

ajratner:
Hey @ThomasPalomares I know of one bug / annoying "feature" where I believe NULL values are dropped from psql arrays... could this be causing any problems perhaps?

ThomasPalomares:
ah maybe ! the error I have is:
ERROR:  malformed array literal: "{"asdf  qwer\tzxcv\n1234",,"NULL","null","\\\\N","N","\\I'm your father","\\\" said Darth Vader.\"","{\\csv in a json\\\": \\\"a","b c","\\\\\\\"","\\\\\\\"","\\\\\\\"line '1'\nbogus","NULL","null","\\\\\\\\N","N","line \\\\\\\"\\\\\\\"2\\\\\\\"\\\\\\\"\\\"","  \\\"foo\\\":123","\n\\\"bar\\\":45.678"," \\\"null\\\": \\\"\\\\\\\\N\\\"}\""}"
so it may be because of that. Do you know if there is a way to modify this in psql or should the ddlib/util.py be modified to take that into account ?

ajratner:
I'm actually not sure... I think I just tried to avoid passing around
arrays of bools in my deepdive apps, come to think of it.  @netj any
thoughts?

netj:
@ajratner Is the bug confusing NULLs in arrays as empty strings?  I gave up this case because I was relying on the `csv` package, but if we're parsing commas ourselves, I think the problem boiled down to a matter of distinguishing `...,,...` from `...,"",...`.

@ThomasPalomares I think the new test is too large involving too many other parts, essentially everything from deepdive: ddlog, compiler, runner, local compute driver etc.

* [ ] Can't we just feed the NastyTSV directly to the `identity.py`, then check whether it outputs the same thing?

```bash
@test "ddlib.util (@tsv_extractor and @returns) works against nasty input" {
    diff -u <(echo "$NastyTSV")  <(python "$BATS_TEST_DIRNAME"/identity.py <<<"$NastyTSV")
}
```
`identity.py` may be even just inlined in the bats file with `-c`.

* [ ] The rest of the content of the .bats file is duplicate and should be removed.

* [ ] Also, the test is not dependent on any database, so let's move it out of `test/postgresql/`, say under `ddlib/test/`.  You can look at other subdir's `env.sh`.

ajratner:
@netj maybe i was only thinking about NULLs being dropped when
array_to_string was used... Thomas and I were going to talk about this
briefly either way-I'm free all weekend!

PR #500, opened 2016-02-22
Fixes spouse example to avoid duplicates of supervised variables

netj:
@ajratner fix to what you pointed out the other day.


ajratner:
LGTM!

PR #506, opened 2016-02-26
Changes to spouse example setup

henryre:
`deepdive do articles` defaults to bash script

CoreNLP requires JDK 1.8 (issue was raised on this too)


doc/example-spouse.md
```diff
@@ -88,12 +88,17 @@ Then we compile our application, as we must do whenever we change `app.ddlog`:
...
-Finally, we tell DeepDive to execute the steps to load the `articles` table:
+Finally, we tell DeepDive to execute the steps to load the `articles` table using the `input/articles.tsv.sh` script. You must have the [full corpus](http://research.signalmedia.co/newsir16/signal-dataset.html) downloaded.
 
 ```bash
 deepdive do articles
 ```
 
+Alternatively, a sample of 1000 articles can be loaded directly by issuing the following command:
+```bash
+deepdive load articles input/articles-1000.tsv.bz2
```
netj: Thanks for the edits!  This is a great!

In fact you need a `deepdive mark done articles` after the `deepdive load` to make it equivalent to `deepdive do articles`.  Could you revise this block as well as the one in quickstart?

doc/example-spouse.md
```diff
@@ -165,7 +170,7 @@ Note that we declare a compound key of `(doc_id, sentence_index)` for each sente
 -->
 
 Next we declare a DDlog function which takes in the `doc_id` and `content` for an article and returns rows conforming to the sentences schema we just declared, using the **user-defined function (UDF)** in `udf/nlp_markup.sh`.
-This UDF is a Bash script which calls [our own wrapper around CoreNLP](https://github.com/HazyResearch/bazaar/tree/master/parser).
+This UDF is a Bash script which calls [our own wrapper around CoreNLP](https://github.com/HazyResearch/bazaar/tree/master/parser). The CoreNLP library requires Java 8 to run.
```
netj: It'd be nice to mention this Java 8 requirement in the first section for preparation/installation as well as in the quickstart, perhaps pointing to this page? http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html

doc/quickstart.md
```diff
@@ -68,18 +68,18 @@ app.ddlog  db.url  deepdive.conf  input/  labeling/  mindbender/  README.md  udf
...
-ln -s articles-1000.tsv.bz2 input/articles.tsv.bz2
+deepdive compile
 ```
+Once it has compiled with no error, you can run the following ```deepdive``` commands.
+
+You can find some of our sampled datasets under `input/`.
+You can also [download the full corpus][corpus], but let's proceed with the one that has 1000 sampled articles. Run the following command to load the sampled articles into DeepDive:
 ```bash
-deepdive do articles
+deepdive load articles input/articles-1000.tsv.bz2
```
netj: Maybe for a quicker quickstart, we could also just tell them to load the preprocessed sentences, not having to mention Java 8, etc.

```bash
deepdive do init/app
deepdive create table sentences
deepdive load sentences input/sentences-1000.tsv.bz2
deepdive mark done sentences
```

(Which doesn't look as simple as before though..)

doc/example-spouse.md
netj: I forgot `deepdive do` was actually doing much more and you need these commands to run before `deepdive load`:

```bash
deepdive do init/app
deepdive create table articles
deepdive load articles input/articles-1000.tsv.bz2
deepdive mark done articles
```

We could come up with a new shorthand for this kind of side-loading operation or revise existing `deepdive load` or `deepdive do`.

netj:
@henryre Thanks for suggesting the edits!  To make it work completely from scratch, we need some revisions.  We may need to add/revise commands to keep things simple though.  Why don't we first make the docs complete as I pointed out, then simplify it in a later PR?

doc/example-spouse.md
henryre: Agreed. I think I had a script doing this and forgot about it. Built-in shorthand is a great idea

henryre:
Sounds good

doc/example-spouse.md
lanphan: What's about using "curl -fsSL git.io/getdeepdive | bash -s _deepdive_runtime_deps" to install jre8 as dependencies, @netj?

henryre:
Whoa what a throwback

netj:
:) Glad we didn't merge this earlier prematurely as it provided strong motivation for improving other parts.

PR #513, opened 2016-03-08
Support for using Torque scheduler with DeepDive

igozali:
This pull request implements a DeepDive extension so that we can execute the processes in the data flow using Torque compiler. Such extensions of this form are called compute-drivers. There are still some cleanup tasks to be performed, but I figured I'd create the PR to get early feedback from all of you! 

Please see`docs/using-torque.md` directory in this branch for more information on how the interface looks like and how to use it. 

Early tests show that there are indeed some usefulness of this extension. Using the Infolab compute cluster, I was able to process the whole spouse articles (signalmedia-1m.jsonl, around 200MB after `grep`ing for wife, spouse, etc.) in an hour by using 25 nodes (as defined by PBS).


igozali:
Reminder in this pull request to discuss `compute-*` filenames; maybe rename it to something better?

lanphan:
Hi @igozali

> Early tests show that there are indeed some usefulness of this extension. Using the Infolab compute cluster, I was able to process the whole spouse articles (signalmedia-1m.jsonl, around 200MB after greping for wife, spouse, etc.) in an hour by using 25 nodes (as defined by PBS).

As I also ran spouse tutorial on 1 machine only, but I try with 2 different OSes:
  * MacOS (laptop, core i7, SSD, 16G RAM):
     + full process (deepdive do probabilities): around 9 hours
     + noted that I did use patch version mkmimo.sh from Jaeho (see #522) + use sed to remove \r, not grep (see #523)

  * Ubuntu (PC, core i7, HDD, 32G RAM):
     + deepdive do has_spouse: around 4h30. "deepdive do probabilities" got problem (see #524)
     + noted that I did use sed to remove \r, not grep (see #523)

So that I'd like want to know your configuration to see how much speedup it is. Would you please tell I know?
Thanks.

igozali:
Hi @lanphan, 

Apologies for the delay, I haven't had time to maintain this pull request until now. 

This branch requires that you have a cluster with more than one machine that is using the [Torque](https://en.wikipedia.org/wiki/TORQUE) scheduler so that you can utilize multiple nodes to run the jobs in the workflow. If you only have one machine, I'm afraid you can't use this feature.

lanphan:
@igozali 
I'd like to ask you about configuration of your cluster (how many RAM, CPU power for each node in your cluster, etc ...)? Because I don't see much speedup comparing to result of my 1 PC only.

igozali:
We have somewhere between 10-30 nodes in the clusters, each having CPU cores in the teens, and sufficient memory. We don't seem to see much speedup due to our DFS configuration, but we haven't verified this yet.

netj:
@igozali Sorry the code has been sitting here forever.  Although many parts can be improved, I think it's in a relatively good shape.  It's probably better to improve it down the road based on what we see from more instances of concrete compute drivers rather than hypothetical ones we imagine now.

igozali:
@netj wow you're actually going over old pull requests! :)

PR #532, opened 2016-04-23
Adds ddlib support for python2 and python3

sandipchatterjee:
* Changes print statements to functions
* Changes relative imports to be explicit (PEP 328)
* Adds \N unicode character escape (raw string)
* Adds makefile tests to be consistent with previous output tests
* Adds some sorts to output methods to make output deterministic -- sorts values
for the features set and in the json.dumps() output to ensure that output is the
same between runs with the same input data (and for the makefile tests)
* Adds new tests within the included Makefile to compare Python3 +/- ddlib
output and Python2/ddlib vs. Python3/ddlib output
* Fixes unit tests in test.py to iterate through generators when necessary

```python
ip-172-30-0-123 python> deepdive env python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from deepdive import *
>>> from ddlib import *
>>> tsv_extractor
<function tsv_extractor at 0x7f0dd2db0c80>
>>> 
ip-172-30-0-123 python> deepdive env python3
Python 3.4.3 (default, Oct 14 2015, 20:28:29) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from deepdive import *
>>> from ddlib import *
>>> tsv_extractor
<function tsv_extractor at 0x7f512f97ae18>
>>> 
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,7 @@
 /examples/tutorial_example/*/experiment-reports/latest
 /examples/spouse_example/data/sentences_dump_large.csv
 /examples/spouse_example/data/spouses.tsv
+
+# python-related
+*.pyc
+__pycache__/
--- a/ddlib/Makefile
+++ b/ddlib/Makefile
@@ -1,6 +1,18 @@
 do:
-	cat test.json | python ./without_ddlib.py > a
-	cat test.json | python ./with_ddlib.py > b
+	cat test.json | python without_ddlib.py > a
+	cat test.json | python with_ddlib.py > b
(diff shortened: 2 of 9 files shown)


netj:
Thanks a lot for the work!  I put a bit more tests around it (turning them into .bats) and merged after fixing another inconsistency in Python3: `map(None, l1, l2, ...)`

PR #535, opened 2016-05-18
Generalized dataflow compiler

netj:
Generalizes the data flow compiler to be applicable to arbitrary workflows, not just DDlog-generated ones.


alldefector:
LGTM

PR #540, opened 2016-06-12
Clean up grounding code to make categorical var inference work

alldefector:
The previous implementation conflates categorical vars with inference rule weight parameters. That made the grounding process a lot more complicated than before we introduced ("dynamic" or "dense") "multinomial factors". However, what we really want is not a new factor type, but rather the ability to support categorical variables in inference rules no matter how rule weight is parameterized (e.g., constant, on category values, or on features).

This change blows away much of the special grounding code for "multinomial factors", simplifying a couple heavy joins.

`make test` passes. The spouse and chunking tests under `test/postgresql` also work. I went through the run logs to make sure the intermediate grounding steps all make sense.

For now, we only support the case where the head is a conjunction of categorical vars. Later we should try to support other factor functions and mixing categorical vars with boolean vars.


alldefector:
buildkit moreutils_0.58 dependency is blocking travis...

compiler/compile-config/compile-config-2.01-grounding
```diff
@@ -73,11 +37,12 @@ def withRedefinitionsWithJoinsForIdColumns:
         ],
         style: "cmd_extractor", cmd: "
         : ${DEEPDIVE_GROUNDING_DIR:=\"$DEEPDIVE_APP\"/run/model/grounding}
-        : ${DEEPDIVE_GROUNDING_DENSE_MULTINOMIAL:=false}
         varPath=\"$DEEPDIVE_GROUNDING_DIR\"/variable/\(.variableName | @sh)
         mkdir -p \"$varPath\"
         cd \"$varPath\"
 
+        deepdive sql \"ANALYZE VERBOSE \(.variablesTable)\"
```
netj: Postgres-specific.  Use [`db-analyze`](https://github.com/HazyResearch/deepdive/blob/master/database/db-driver/postgresql/db-analyze) instead.  Adding `VERBOSE` to postgres' may be a good idea.

compiler/compile-config/compile-config-2.01-grounding
```diff
@@ -454,65 +413,10 @@ def withRedefinitionsWithJoinsForIdColumns:
...
-                                        , { table: "i.\($columnIdForThisVar)", column: . } ] } ) ] }
-                        ) ]
-                    } | asSql | asPrettySqlArg)
-            fi
-            " else # factors over boolean variables
             "
             # find distinct weights for the factors into a separate table
             deepdive create table \(.weightsTable | @sh) as \($weightsMaterializeSqlArg)
-            " end)
+            ")
```
netj: For readability let's inline `$weightsMaterializeSqlArg` since it's no longer repeated.

compiler/compile-config/compile-config-2.01-grounding
```diff
@@ -752,81 +544,39 @@ def withRedefinitionsWithJoinsForIdColumns:
...
+                        "\(.columnPrefix)\(deepdiveVariableInternalLabelColumn)" | asSqlIdent
+                        ] | join(", ")) as $valueOrder
+                    |
                     { SELECT:
                         [ ( .function_.variables[]
                         | { table: "f", column: .columnId } )
                         , { alias: "num_weights", expr: "COUNT(w.wid)" }
-                        , { alias: "weight_ids" , expr: "ARRAY_AGG(w.wid ORDER BY w.wid)" }
+                        , ( .function_.variables[]
+                        | { alias: "c\(.ordinal)_ids",
```
netj: Maybe add a comment that these two columns are parallel arrays?

compiler/compile-config/compile-config-2.01-grounding
```diff
@@ -427,16 +375,27 @@ def withRedefinitionsWithJoinsForIdColumns:
 
             # materialize factors using user input_query that pulls in assigned ids to involved variables
             deepdive create table \(.factorsTable | @sh) as \(
-                # allow .input_query to refer to the assigned variable ids by redefining the variable table names with SQL CTE
+                # Not using DISTINCT: User decides if they want to have duplicate factors
```
netj: Note that `factorsTable` can be a view.  Materializing this can happen in the `factors/*/dump` process.  weightsTable below can come directly from the `input_query` instead.  Change can be simple as `deepdive create view ... as ...` or we could move and plug the query into the dump process.

In general, only the ones up for sequential id assignment have to be materialized as table.  If user's `input_query` is costly we could materialize that instead, but I think factor input_query is expected to generate bloat by design since the result is supposed to be projected and group'ed/distinct'ed, so maybe that's a bad idea.

netj:
Looks much clearer!  Glad to see another example of less code making things more general.  I forgot what bug is still there?

alldefector:
The biased coin example? The deepdive.conf code stopped working because of the var table change...

Will address the comments soon!

compiler/compile-config/compile-config-2.01-grounding
alldefector: If it's a view and involves expensive joins, we would evaluate it twice. I'm leaving it as a table for now.

netj: Okay.  I have a feeling it'll waste time&space to materialize what's going to be immediately projected out, but we can do this comparison more rigorously and decide later.

compiler/compile-config/compile-config-2.01-grounding
alldefector: done

compiler/compile-config/compile-config-2.01-grounding
alldefector: done

compiler/compile-config/compile-config-2.01-grounding
alldefector: done

alldefector:
Travis seems to always fail because the chunking example times out (10min):
`No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.`

alldefector:
@netj PTAL. Also, is it possible to increase the travis timeout limit? (Could also try reducing number of docs in chunking test.)

netj:
@alldefector Yes I was testing things myself, and it took a while on my new Mac.  I don't think we can increase Travis timeout nor would it be a good thing to have to wait for an hour per test, so let's reduce the data size.

netj:
I think we can shrink things to 20% or less (just on Travis or by default) and lower the F score bar for chunking (getting ~0.65-0.7 for that size) since we're mainly interested in whether it works end to end.
```
export SUBSAMPLE_NUM_WORDS_TRAIN=1000 SUBSAMPLE_NUM_WORDS_TEST=200
```

I'm also noticing that spouse example now takes significantly longer than before on my Mac.  postgres stuck at materializing the JOIN with dd_variable_* table.

alldefector:
Yeah, those joins from reuse_weight stage of the spouse example never finish on my macOS either... I'm trying reducing the chunking numbers.

alldefector:
In the spouse test, the "postgresql spouse example reuse weights" step, there are only 3688 variables, but the big join in https://github.com/HazyResearch/deepdive/issues/541 could not finish even after 2HR... Maybe we should try to change the key from two columns to one.

PR #542, opened 2016-06-18
Add timestamp conversion from Postgres dump to JSON

shahin:
The `pgtsv_to_json` module is the fallback for database systems that don't support `to_json()`. It's called when I run on Greenplum, but I have timestamps in my table. Without this commit, the module terminates on a ValueError. With this commit, timestamp input columns are included in the JSON output.

### DIFF
--- a/database/pgtsv_to_json
+++ b/database/pgtsv_to_json
@@ -7,17 +7,19 @@ from datetime import datetime
 
 def timestamp(timestamp_str):
     """Given a timestamp string, return a timestamp string in ISO 8601 format to emulate
-    Postgres 9.5's to_json timestamp formatting. The given timestamp is assumed UTC.
+    Postgres 9.5's to_json timestamp formatting.
 
-    Time zones are not supported. http://bugs.python.org/issue6641
+    This supports the `timestamp without time zone` PostgreSQL type.
+
+    Time zone offsets are not supported. http://bugs.python.org/issue6641
 
     Examples:
 
         >>> timestamp('2016-06-17 20:10:38')
-        '2016-06-17T20:10:38+00:00'
+        '2016-06-17T20:10:38'
 
         >>> timestamp('2016-06-17 20:10:37.9293')
-        '2016-06-17T20:10:37.929300+00:00'
+        '2016-06-17T20:10:37.929300'
 
     """
 
@@ -25,7 +27,7 @@ def timestamp(timestamp_str):
         parsed = datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S.%f')
     except ValueError:
         parsed = datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S')
-    return parsed.isoformat() + '+00:00'
+    return parsed.isoformat()
 
 # given a column type, returns a function that takes a string input
 # and output with the correct type
@@ -48,7 +50,7 @@ def convert_type_func(ty, ty_rest = ""):
       return convert_other_array
   else: # non-array, must be primitive type
     normalized_type_name = {
-        "timestamp": "timestamp",
+        "timestamp without time zone": "timestamp",
         "integer" : "int",
         "bigint"  : "int",
         "double"  : "float",
@@ -64,8 +66,8 @@ def convert_type_func(ty, ty_rest = ""):
         }
     # find the convert function with normalized type name
     ty = ty.lower()
-    if ty.startswith('timestamp'):
-      ty = 'timestamp'
+    if re.match('timestamp(\(\d\))? without time zone', ty):
+      ty = 'timestamp without time zone'
     if ty in normalized_type_name:
       ty = normalized_type_name[ty]
     if ty in convert_for_primitive_types:
--- a/database/pgtsv_to_json
+++ b/database/pgtsv_to_json
@@ -3,6 +3,29 @@
 # Usage: pgtsv_to_json COLUMN_NAME1:TYPE1 COLUMN_NAME2:TYPE2 ...
 
 import json, sys, csv, re
+from datetime import datetime
+
+def timestamp(timestamp_str):
+    """Given a timestamp string, return a timestamp string in ISO 8601 format to emulate
+    Postgres 9.5's to_json timestamp formatting. The given timestamp is assumed UTC.
+
+    Time zones are not supported. http://bugs.python.org/issue6641
+
+    Examples:
+
+        >>> timestamp('2016-06-17 20:10:38')
+        '2016-06-17T20:10:38+00:00'
+
+        >>> timestamp('2016-06-17 20:10:37.9293')
+        '2016-06-17T20:10:37.929300+00:00'
+
+    """
+
+    try:
+        parsed = datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S.%f')
+    except ValueError:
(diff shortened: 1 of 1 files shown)


database/pgtsv_to_json
```diff
@@ -25,20 +48,24 @@ def convert_type_func(ty, ty_rest = ""):
       return convert_other_array
   else: # non-array, must be primitive type
     normalized_type_name = {
+        "timestamp": "timestamp",
```
netj: This seems unnecessary, but fine to keep.

database/pgtsv_to_json
```diff
@@ -3,6 +3,29 @@
...
+        >>> timestamp('2016-06-17 20:10:37.9293')
+        '2016-06-17T20:10:37.929300+00:00'
+
+    """
+
+    try:
+        parsed = datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S.%f')
+    except ValueError:
+        parsed = datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S')
+    return parsed.isoformat() + '+00:00'
```
netj: It'd be nice to default to local timezone but sounds like it's not easy to get it or is it just the lack of support for `%:z`?  Does GP store all timestamp in UTC btw?

shahin: Timezone-aware dates are stored internally in UTC, but they're output in whichever time zone is set for the session. This can be (a) the system time zone, (b) the time zone set in postgresql.conf, (c) the current time zone setting for the psql session, or (d) [something else](https://www.postgresql.org/docs/8.3/static/datatype-datetime.html#DATATYPE-TIMEZONES). Given that, defaulting to "local" time seems potentially convenient but very easily incorrect (it's not really clear what "local" means in this context).

This code will raise on any timestamp with a timezone, so it's not handling them incorrectly -- it's refusing to handle them at all.

Now that you mention it, it is pretty odd to output a UTC zone offset in this case. It shouldn't output a zone offset at all. I can remove the `+00:00` on output. How does that sound?

shahin: The `pgtsv_to_json` module just takes text over STDIN and isn't aware of any database sessions. Inspecting a separate psql process for state seems brittle, especially since there's no guarantee one exists.

I'm getting the impression that you'd like this module to handle time zone offsets, which is totally reasonable. I like the original module's design, so my preference is to avoid:

- changes to the `pgtsv_to_json` interface,
- changes outside this Python module,
- assumptions about the state of a psql session.

Two options are:

1. Parse and output time zone offsets where they exist on the input, and output no time zone offset where they don't, or
2. Raise a `ValueError` where time zone offsets exist on the input, and output no time zone offsets where they don't.

Either way there's no loss of accuracy, and I'm happy to go in either direction. What do you think?

alldefector: 1 sounds better. But we are probably over-engineering here. I'd simply output verbatim whatever psql feeds us. Chances are the downstream app would actually prefer verbatim. For the current main consumer of ES, it supports a variety of formats already: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html

netj:
I meant we can set the time zone in the psql that feeds input to the script such that the UTC offset we add is going to be always correct. The script doesn't have be general since pgtsv_to_json is really just a shim for working around a limitation in some db-drivers and not intended for general use outside of it.

shahin:
@alldefector Good suggestion. I tried passing the timestamp string through unchanged, but `jq` can't handle that particular format. This suggests to me that other JSON tools won't either, even if ES does.

@netj Our timestamps are time zone naive and don't have offsets (they use Postgres' `timestamp without time zone` type), so there is no time zone setting in psql that can result in a correct offset down the line. We could change these columns to `timestamp with time zone` but now we're asking the application to support time zones for a really obscure technical reason.

I pushed 0a794ff in an attempt to resolve this. Before this change, `pgtsv_to_json` didn't support any timestamps. After this change, `pgtsv_to_json` explicitly supports the `timestamp without time zone` type without qualification. Looks to me like `jq` and Elastic support the output without changes elsewhere.

Please let me know if you think it produces inaccurate results, or if it doesn't support a feature that we need!

netj:
I don't understand why anyone should ever want to use timestamp without time zone (it doesn't make sense as it's a reference to an absolute point in time and not subject to interpretation). http://stackoverflow.com/a/6158432

It's fine to use any time zone in the JSON output as long as we are consistent with the data source. Since pgtsv to json is used only once in our codebase https://github.com/HazyResearch/deepdive/search?utf8=✓&q=pgtsv_to_json, we can just put the following line before the COPY statement to ensure we get timestamps serialized in a certain tz
```
SET timezone = 'UTC';
```
Not sure why pgtsv to json needs any change from the first version. This is a problem in the data source not the data encoder. 

BTw JSON defines one and only one standard datetime format: ISO8601 which is also used by XML Schema, etc. so i agree we shouldn't produce some nonstandard format. It's no longer JSON if we try to sneak in some weird date time format.

shahin:
@netj I agree, it's weird and certainly incorrect to leave out time zone info! Unfortunately adding time zone support to our app is way beyond the scope of my current task (moving specific processing from Postgres to Greenplum).

I don't think I made the motivation clear in my initial PR, and my original patch was definitely flawed. Sorry about that! The motivation for this PR is twofold:

1. `pgtsv_to_json` [terminates on a ValueError](https://github.com/HazyResearch/deepdive/blob/c303c9c7ac9d130ba60dcaab52e74c0d1937ebe4/database/pgtsv_to_json#L47) when dumping timestamps of any kind, and
2. `jq` crashes when we send it psql-formatted timestamp strings, which are missing the "T" character and as you pointed out are not valid ISO 8601.

Seems like setting psql's time zone to UTC for this one dump doesn't fix those problems, but does make our timestamps incorrect until we support time zones from the very beginning of the pipeline. As it stands the data source (PostgreSQL/Greenplum) doesn't contain time zone information, and no combination of psql runtime directives can create that information out of thin air.

Would you be open to my approach in 0a794ff *along with* a promise to open a ticket for proper time zone handling in downstream apps? I'd love to fix that at some point.

alldefector:
@shahin Related: The missing 'T' issue is a bug in postgres's `to_json` that was fixed only in 9.5...
http://postgresql.nabble.com/BUG-12578-row-to-json-and-to-json-add-T-in-timestamp-field-td5834396.html

netj:
Ok make sense now. Keeping a way to pass thru incomplete data may be something useful too anyway.

database/pgtsv_to_json
```diff
@@ -3,6 +3,31 @@
...
+
+        >>> timestamp('2016-06-17 20:10:37.9293')
+        '2016-06-17T20:10:37.929300'
+
+    """
+
+    try:
+        parsed = datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S.%f')
+    except ValueError:
+        parsed = datetime.strptime(timestamp_str, '%Y-%m-%d %H:%M:%S')
```
alldefector: If this fails too, can we just return the input string?

alldefector:
LGTM

database/pgtsv_to_json
shahin: Added in f5dc3d3

shahin:
@alldefector Looks like CI is failing, but I don't think it's caused by this diff. Can I merge and push on the command line anyway? Should I open a ticket for fixing CI here?

PR #543, opened 2016-06-19
Var values are int with refactored sampler

alldefector:
With https://github.com/HazyResearch/sampler/pull/30 , sampler now expects variable init_values to be INT (instead of DOUBLE).


netj:
LGTM

PR #544, opened 2016-06-19
Purge "multinomial"

alldefector:


compiler/compile-config/compile-config-2.01-grounding
```diff
@@ -378,15 +378,15 @@ def factorWeightDescriptionSqlExpr:
                 # Not using DISTINCT: User decides if they want to have duplicate factors
                 { SELECT:
                     [ ( .function_.variables[] | { table: "F", column: .columnId } )
-                    # TODO XXXX: assuming so-called "multinomial" factors' heads have categorical vars only
-                    , ( .function_ | select(.isMultinomial) | .variables[] |
+                    # TODO XXXX: assuming categorical factors' heads have categorical vars only
```
netj: I think three-X is the convention: `XXX`.  Editors and IDEs recognize them and highlight just like `TODO` and `FIXME`.  Not sure where they came from but pretty useful to flag comments.

netj:
Looks great!  Should we also get rid of the categories column in the weights view here, I remember you flagged as unnecessary in a previous commit?

compiler/compile-config/compile-config-2.01-grounding
alldefector: I thought more Xs could get more attention :)

alldefector:
I assume you meant dd_graph_weights.categories? Removing now.

PR #549, opened 2016-06-21
Fix more unicode errors

shahin:
This diff corrects:

1. Incorrect parsing of Postgres' `COPY FROM` output for strings containing unicode, and
2. Incorrect processing of Python escape sequences like `\n` in unicode strings


alldefector:
Let's just port this to python3.

shahin:
Closing in favor of #551

PR #555, opened 2016-07-01
Balance example to show badness of linear semantics

alldefector:


examples/balance/input/yes.tsv
```diff
@@ -0,0 +1,1000 @@
+1
```
netj: You can replace this with a line in `input/yes.tsv.sh`:
```
seq 1000
```
and `chmod +x input/yes.tsv.sh`.  Same thing applies to `no.tsv`.

alldefector: done

netj:
Nice.  Maybe we should use a more descriptive name? `imbalance` or `bad_linear_semantics`?

alldefector:
@netj The naming is optimized for the future, i.e., when we manage to implement the RATIO semantics, it's also the time we achieve balance :)

examples/balance/app.ddlog
```diff
@@ -0,0 +1,12 @@
+yes (vote_id bigint).
+no  (vote_id bigint).
+
+whatever? (
+    candidate bigint
+).
+
+@weight(1)
+whatever(p) :- yes(_).
```
netj: To match the paper more closely, I think this should be an imply factor rather than a unary (also turning the two others into variables).
```
@weight(1)
yes(_) => whatever(p) :- TRUE.
```
In the paper we overloaded the datalog head/body syntax with implication, which later has been disambiguated into a separate syntax.  That gives users precise control of mapping data to the model.  (I know, the `:- TRUE` part looks ugly and shouldn't be mandatory)

alldefector: Grounding would fail with that syntax... I think currently we require all relations in the inference rule head to be a variable relation.

netj: Right. So they also need the question marks in the schema and can be given labels to act as evidence.

alldefector: OK, I don't think it's necessary to make `yes` / `no` variables or use IMPLY. The ratio semantics concern the counts of "positive" and "negative" factors.

PR #558, opened 2016-07-06
Usability tweaks and many bug fixes

netj:
Notably:
- Faster `deepdive compile`
  - replacing HOCON parsing in Scala with pyhocon
  - single ddlog compiler (JVM) boot up
- Automatic `deepdive compile` by `deepdive do`-like commands
- Greenplum driver checking `psql` client
- DimmWitted NUMA bug fixes
- Bundled Python virtualenv (`extern/bundled/python-virtualenv/requirements.txt`)
- Other small yet nasty bugs that used to waste lots of time from many users


extern/bundled/python-virtualenv/install.sh
```diff
@@ -0,0 +1,38 @@
...
+set -euo pipefail -x
+
+unset PYTHONPATH  # existing PYTHONPATH can interfere
+
+( # install a Python virtualenv
+if ! type virtualenv &>/dev/null; then
+    # if virtualenv isn't available, install it locally along with pip
+    [[ -e bootstrap/virtualenv.py ]] || {
+        curl -RLO https://bootstrap.pypa.io/get-pip.py &&
+        python get-pip.py virtualenv --ignore-installed --target bootstrap
```
alldefector: Could `python` here be python3? Should we bundle `python` as well?

compiler/app-has-nothing-to-compile
```diff
@@ -0,0 +1,24 @@
...
+export DEEPDIVE_APP
+cd "$DEEPDIVE_APP"
+
+# source files to check
+set -- app.ddlog deepdive.conf schema.json
+
+# check if each source file is newer than last compiled result
+num_updated=0
+for src; do
+    [[ "$src" -nt "$DEEPDIVE_APP"/run/compiled/config.json ]] || continue
```
shahin: seems like this is what `make` does but I see that `deepdive-compile` is a custom script instead a makefile. Is this just a preference or does it reflect an incompatibility between our build and `make`?

netj: Good point. Large portion of it can be rewritten into a makefile but there's a little extra done so we'll still need some custom steps pre/post `make`.

extern/bundled/python-virtualenv/install.sh
netj: I guess it'll follow the system default at build time which may be undesirable. `python` also seems to get bundled as part of virtualenv but it's not exposed and only used for the pyhocon command. My intention for the bundled virtualenv was to serve only internal components but we might want to find a nice way to provide a default set of packages to the user. What's lacking here, I've been increasingly worried about, is a clear separation between the environment user's codes run in and deepdive's internal components do. We probably should try to not pollute user's env and document what exactly gets added to it, e.g., ddlib, psycopg2. Ideally users should be able to run deepdive in another virtualenv and their udfs shouldn't get interfered by the bundled virtualenv. Curious to hear any thoughts along the line!

alldefector: Agree that it's probably a bad idea to pollute the runtime user's env. For internal use, should we just stick with `python3` (better unicode handling at least, also no ambiguity)? Do we ever need to bundle ddlib (assuming it's the ancient feature lib)?

netj: I'd also like to default to python3 but need to check if virtualenv is standalone or needs a host python3.  ddlib provides some crucial decorators for UDFs (`@tsv_extractor` parsing/formatting PG TSV and `@tsj_extractor` for TSJ in the future) and the new automatic feature lib is not here yet, so yes users want/need ddlib.

alldefector:
LGTM!

PR #559, opened 2016-07-07
Add dependency for perl-Time-HiRes

clfarron4:
Following error is raised when running deepdive run_deepdive_tests

    `DEEPDIVE_COMPILE_INPUT_JSON= deepdive compile || false' failed
       Can't locate Time/HiRes.pm in @INC (you may need to install the Time::HiRes module) (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .)

### DIFF
--- a/util/install/install.RedHat.sh
+++ b/util/install/install.RedHat.sh
@@ -59,6 +59,7 @@ install__deepdive_runtime_deps() {
         java
         gnuplot
         libtool-ltdl # for graphviz
+        perl-Time-HiRes
     )
     sudo yum install -y "${runtime_deps[@]}"
     sudo localedef -f UTF-8 -i en_US en_US.UTF-8


netj:
@clfarron4 Thanks for the fix!  I think a runtime dependency that DD bundles requires it (moreutils `ts`), which seems to be included in modern Perl distributions (at least since 5.16).  Out of curiosity could you share your `cat /etc/redhat-release` and `perl -v`?

clfarron4:
Apologies for the delay. Here is the information requested.

    [claire@localhost ~]$ cat /etc/redhat-release
    Fedora release 24 (Twenty Four)
    [claire@localhost ~]$ perl -v
    
    This is perl 5, version 22, subversion 2 (v5.22.2) built for x86_64-linux-thread-multi

PR #561, opened 2016-07-12
Fix categorical grounding: GROUP_BY non_category_weight_cols

alldefector:


compiler/ddlog
```diff
@@ -1 +1 @@
-Subproject commit 22f4fdb40dc2e165e501affd9091267f1acfe200
+Subproject commit 091296cd6e04fc21e9ec384eba6f2b5de7ec90c6
```
netj: Can you update ddlog to master after the merge that just happened?

test/postgresql/chunking_example.bats
```diff
@@ -4,10 +4,10 @@
 . "$BATS_TEST_DIRNAME"/env.sh >&2
 
 : ${CHUNKING_TEST_NUM_WORDS_TRAIN:=1000} ${CHUNKING_TEST_NUM_WORDS_TEST:=200}
-: ${CHUNKING_TEST_MIN_F1SCORE:=60}
+: ${CHUNKING_TEST_MIN_F1SCORE:=80}
 
 : ${CHUNKING_TEST_REUSE_NUM_WORDS_TEST:=200}
-: ${CHUNKING_TEST_REUSE_MIN_F1SCORE:=60}
+: ${CHUNKING_TEST_REUSE_MIN_F1SCORE:=80}
```
netj: Interesting.. the skip chains were actually hurting the quality?

netj:
LGTM

test/postgresql/chunking_example.bats
alldefector: It was PHONY! It was a mesh (<i, j> for all 0 <= i < j < N), not skip chain! Keeping that rule would drag the F1 to 53. Removing it gets us F1 = 87.

PR #562, opened 2016-07-16
Continuous feature support

alldefector:
Related:
https://github.com/HazyResearch/ddlog/pull/105
https://github.com/HazyResearch/sampler/pull/42

Now we can use the `@value(field)` annotation on an inference rule to pass real values as features (multiplier of potential functions) as opposed to the implicit {0,1} for discrete features.

Example(s) coming soon.


alldefector:
@netj PTAL

PR #564, opened 2016-07-16
Build and test inside Docker containers

netj:
Docker for Mac beta finally made containers usable on Mac running them natively.  Debugging broken Travis builds (esp. timeout errors) was wasting too much time.  So here's a step forward embracing containers to capture and reproduce every build step.  On the host, only Docker needs to be installed. (and bash and jq and maybe a few more utilities... but nothing crazy that we can remove down the road.)

Note that a single `Dockerfile` building everything after installing everything from scratch is infeasible due to its high latency for iterative development and even for just running repetitive CI builds.  So, we take a two-tier'ed approach:

1. First, there is a typical `Dockerfile` that takes the source tree and builds an image from scratch, we call the "master" image.
    * `docker build` or `DockerBuild/update-master-image-from-scratch` can be used

2. Second, there are a few help scripts supporting incremental build and test workflow on top of the master image:
    * `DockerBuild/build-in-container`
    * `DockerBuild/test-in-container*`
    * `DockerBuild/update-master-image`

More details are documented in [DockerBuild/README.md](https://github.com/HazyResearch/deepdive/tree/a700a5651bec8060ad1868501fad6fd953b9f931/DockerBuild#readme).  Everything under DockerBuild/ is general, so we should consider submodule'ing it (or just copy/paste) to other submodules (dimmwitted, ddlog) to containerize their builds.

Travis now uses DockerBuild/ and pushes builds and test results to Docker Hub, so we can easily reproduce and investigate build failures.  Built images appear here:
* https://hub.docker.com/r/netj/deepdive-build/tags/


Dockerfile
```diff
@@ -0,0 +1,43 @@
+# Dockerfile to build and test DeepDive inside a container
+#
+# `make build-in-container test-in-container` uses master image built by this.
+# `util/build/docker/` contains utilities relevant to this.
+FROM ubuntu
```
alldefector: Maybe specify a version? say 16.04

DockerBuild/test-in-container-postgres
```diff
@@ -0,0 +1,43 @@
...
+# wait until postgres starts
+postgres-has-started() {
+    docker run --rm --name "$DOCKER_PGNAME.check" \
+        --link "$DOCKER_PGNAME" \
+        --env PGPASSWORD="$POSTGRES_PASSWORD" \
+        postgres \
+        psql -h "$DOCKER_PGNAME" -U "$USER" -l
+}
+sleep $POSTGRES_START_TIME  # give it some initial time to start
+while ! postgres-has-started; do sleep 1.$RANDOM; done
```
alldefector: Not sure if it's worth it for now: if we need more services as fixture in the future, we may want to try docker compose.

alldefector:
LGTM!

DockerBuild/test-in-container-postgres
netj: Sure we should definitely switch to the right way like docker compose soon.  I didn't have chance to play with it yet and didn't want to complicate things without fully understanding it, which will take some time.  Btw Greenplum tests should be fixed and included in Travis first (it's not working yet as the clients are missing in the build image).

Dockerfile
netj: There are pros and cons.  I think the source code should be portable, even for non-ubuntu, so I'd vote against fixing it.  Any compelling reason to do so?

alldefector: I don't know how "ubuntu" is resolved -- maybe it's resolved to "latest" so in October it'll be ubuntu:16.10 right after the release. But if during the next 18 months, production envs stick to 16.04 (which is the most likely scenario), then Travis would be testing against an env different from prod during that time. When DD breaks on 16.04 but not on 16.10, Travis wouldn't catch that.

netj: This Dockerfile and DockerBuild/ scripts are for development.  For production and normal users, I think we should have a separate set of Dockerfiles that just installs the built DD binary releases with their runtime dependencies.  For that we can/should certainly fix the ubuntu version.  However, if we're running everything containerized, perhaps the host OS distro doesn't even matter.  We could just mount the DD app and run it to populate another database container.  However, yes for standalone releases, we may want to test it against various flavors of distros and versions, or just explicitly target few of them.

Justification for separate production images: The deepdive-build image is now ~4-5GiB.  However, given that the release binary is ~100MiB compressed and as the runtime dependencies are going to be much smaller than build deps, I'd say the production image can easily go under 1GiB.

netj: And yes, `ubuntu` is `ubuntu:latest` and I have to check if their policy is to strictly follow LTS or literally any latest release. (I hope the former but may not be the case)

PR #565, opened 2016-07-19
TAB-separated JSONs

netj:
This PR implements TSJ to be used as DeepDive's primary SerDe format, replacing TSV.

TSV is too hard/heavy to parse correctly, and JSON object-per-line (sometimes called jsonl) has too much overhead repeating field names on every single line even though the schema is fixed.  TSJ is the best of both worlds:
1. Parsing (decoding/deserialization) is super easy: split each input line by TAB, then parse each field with universally available JSON parser.
2. Formatting (encoding/serialization) is super easy: encode each output value in JSON, again universally available, then join with TAB to produce a line.
3. Extra JSON encoding overhead is negligible compared to TSV.

It took some effort, but as opposed to (PG) TSV, which has been creating ever more burden on every participants across the data flow to handle the format correctly, TSJ support by DeepDive's db-drivers removes the burden once and for all.

Here's a list of things to highlight:
1. `deepdive sql eval SQL format=tsj`
2. `@tsj_extractor` decorator for UDFs.
3. `deepdive load foo input/foo.tsj`
4. `deepdive unload foo input/foo.tsj` with Perl/Python implementations (Python psycopg2+ujson is the default as it showed much less overhead than parsing PG TSV and mapping it to TSJ with Perl)
5. `psycopg2` added to the DD-bundled Python virtualenv
6.  postgresql 9.5 bundled and used unless user has one already


database/db-driver/postgresql/db-query-tsj
```diff
@@ -0,0 +1,73 @@
...
+TYPES_DATETIME_PATTERNS = [re.compile(regex) for regex in (
+        r'date',
+        r'interval(?:|.+)',
+        r'timestamp\s*(?:\(\d\))?(?:\s*with(?:out)?\s+time\s+zone)?',
+        )]
+
+# generate code to output columns of each row based on their types
+# See: http://lucumr.pocoo.org/2011/2/1/exec-in-python/
+stmts = []
+for i,ty in enumerate(column_types):
```
shahin: Should this be wrapped in a testable function?

database/db-driver/postgresql/db-query-tsj
```diff
@@ -0,0 +1,73 @@
...
+exec compile("def write_tsj_output(columns):\n" + "\n".join("  " + s for s in stmts),
+        "%s codegen" % sys.argv[0], "exec")
+
+# tell psycopg2 to actually skip parsing any JSON
+psycopg2.extras.register_default_json(loads=lambda x: x)
+# XXX or use ujson.loads if you really must parse it
+#psycopg2.extras.register_default_json(loads=ujson.loads)
+
+# use pyscopg2 to access the database
+conn = psycopg2.connect(
```
shahin: This will try to connect to some database when the module is imported. I know that we're running it as a script in deepdive, but this makes it harder to test, won't play nice with pdb, and makes it difficult to compose with other Python.

We can have the best of both worlds, though, if we throw all the connection stuff into an `if __name__ == '__main__'`. Would that work here?

database/db-driver/postgresql/db-query-tsj
```diff
@@ -0,0 +1,73 @@
...
+# $ db-query-tsj SQL COLUMN_TYPE...
+##
+import psycopg2, psycopg2.extras, ujson
+import sys, subprocess, os, re
+
+# get arguments from command-line
+def usage(*msg):
+    err = subprocess.call(["usage", sys.argv[0]] + list(msg))
+    if err != 0: sys.exit(err)
+if not len(sys.argv) > 1: usage("Missing SQL")
```
shahin: I'm thinking argument checks should go in `if __name__ == __main__` in case someone needs to import this instead of running from the command line.

ddlib/ddlib/util.py
```diff
@@ -371,3 +379,45 @@ def tsv_extractor(generator):
...
+  """
+  reload(sys).setdefaultencoding("utf8")  # to avoid UnicodeEncodeError of JSON values during conversion by str()
+
+  input_format, output_format = get_generator_format(generator)
+  input_names  = [name for name,t in input_format]
+  num_input_values = len(input_format)
+  num_input_splits = num_input_values - 1
+  num_output_values = len(output_format)
+
+  def parse_json(column_index, json_value):
```
shahin: Can we make this a top-level, testable/importable function?

shahin:
This is very cool. Can you include stats on wall time and space savings for TSJ vs JSON, and TSJ vs TSV?

I know that you wrote this specifically for our Deepdive use cases, but serde isn't a Deepdive-specific function. Can we break your core TSJ parsing/serialization code into a separate Python package? This will make it more testable, transparent and composable. If it's out of scope here, feel free to open a ticket so that we can get to it later.

I see that `DEEPDIVE_USE_TSV2TSJ` lets us switch from the Python lib to a Perl version. When should we do this? The Perl stuff should probably be pulled out into its own installable module, too.

database/db-driver/postgresql/db-query-tsj
netj: Good point.  My goal was to keep code minimal without any frills just for implementing this executable, but I agree this piece can be a generally useful Python module.  I'm thinking of moving this into somewhere like ddlib.pgtsj, have it installed in our virtualenv, and have this script call the main after doing a bit of argument parsing.  .pyc's can be used that way although I doubt it'd make any big difference for this one.  How does that sound?

Btw Is there a CLI generator you recommend?  It seems pip, virtualenv, etc. were all following a convention and their script seemed generated.

shahin: Yeah, a separate package that we install in our virtualenv sounds great. I agree that we don't need to do anything special w/ .pyc's here.

When you say "CLI generator" I think of two different things:

1. For exposing executable commands from an installed Python package, consider using `entry_points`. Here's [a general overview](https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/) and here's [an example in memex](https://github.com/LatticeData/memex/tree/71b8ccb0730392c7ae442b44413eb6dae6ec3c69/queryjson). 

2. For parsing arguments, [`argparse`](https://docs.python.org/3/library/argparse.html) is the most common and straightfoward. When I have a really complex interface with lots of options I tend to use [`docopt`](https://github.com/docopt/docopt/blob/master/examples/arguments_example.py).

Let me know if I misunderstood and you were actually asking about something else.

netj:
@shahin Thanks for taking a look!

* Sorry those numbers were buried in the commit messages.  Let me copy them here for everyone.  Below are three commits showing how Python implementation's overhead went from 25-30x down to 1.5-3x.

* I'd say these measurements aren't DeepDive-specific as the db-drivers just called the TSJ pieces (`db-query-tsj` and `tsv2tsj.pl`) and the rest was just direct interaction with postgres and them.

* We can throw away the Perl impl., but it's just kept here as it can be useful for converting existing PGTSV to TSJ format.  However, I doubt it'll be more useful to anyone by giving a name usable in Perl like `use DeepDive::PGTSJ qw(tsv2tsj); ...`.  Command-line is already too simple:
    ```
deepdive env tsv2tsj <old.tsv >new.tsj
``` 

* Nothing was measured space-wise as it's quite obvious.  TSJ had a few % increase over TSV on a dataset with many text arrays (per-token NLP markups in the sentences table).  But you can cook any number you want with a different dataset and schema.

----

1. TSJ unloading throughput experimentation (c62a2113a6618c000144afb144ea796a25f98edb)
> I was hopeful for psycopg2 + ujson to deliver some throughput, but unfortunately there’s even more overhead..
> obviously while unpacking the values from PG and serializing into JSON.
> 
> With simpler table (two text columns with large content) it’s quite good.
> For unloading the articles table in our spouse example repeated 100 times (~423MiB, 111k rows):
> 1. PG TSV shows 204MiB/s or 55k/s
> 2. psycopg2 + ujson in Python single threaded gives ~142MiB/s or 37k/s
> 3. PG TSV to TSJ (w/ Perl regexes) parallel gives ~40MiB/s or 10k/s
> 4. PG TSV to TSJ (w/ Perl regexes) single threaded gives ~16MiB/s or 4.2k/s
> 
> However it got 25x slower (from 1.5x) with more complex table (9 columns: text, int, json arrays).
> For unloading the sentences table in our spouse example repeated 100 times (~3GiB, 3.11M lines):
> 1. PG TSV shows 320MiB/s or 314k/s
> 2. PG TSV to TSJ (w/ Perl regexes) parallel gives ~56MiB/s or 55k/s
> 3. PG TSV to TSJ (w/ Perl regexes) single threaded gives ~22MiB/s or 22k/s
> 4. psycopg2 + ujson in Python single threaded gives ~13MiB/s or 13k/s
> 
> Rought notes on how to reproduce:
> 
>     export PATH="$PWD"/dist/stage/bin:"$PATH"
>     export DEEPDIVE_DB_URL=postgresql:///deepdive_tsj_unload_throughput
> 
>     : prep input
>     cd examples/spouse
>     deepdive db init
>     deepdive load articles  input/articles-1000.tsj.bz2
>     deepdive load sentences input/sentences-1000.tsj.bz2
>     cd -
>     deepdive create view a100 as "SELECT * FROM articles  $(perl -e 'print " UNION ALL SELECT * FROM articles " x 100')"
>     deepdive create view s100 as "SELECT * FROM sentences $(perl -e 'print " UNION ALL SELECT * FROM sentences" x 100')"
> 
>     : 1. PG TSV
>     deepdive unload a100 a.tsv
> 
>     : 2. PG TSV to TSJ (w/ Perl regexes) parallel (8-way)
>     DEEPDIVE_USE_TSV2TSJ_NPROCS=8  deepdive unload a100 a.tsj
> 
>     : 3. PG TSV to TSJ (w/ Perl regexes) single threaded
>     DEEPDIVE_USE_TSV2TSJ_NPROCS=1  deepdive unload a100 a.tsj
> 
>     : 4. psycopg2 + ujson in Python single threaded
>     DEEPDIVE_USE_TSV2TSJ=false     deepdive unload a100 a.tsj
> 
>     : repeat with s100 s.tsv and s.tsj

2. Improves db-query-tsj throughput (49230ee40009cff709aa76966cdd72b713daeb16)
> Using ujson.loads for psycopg's json typecaster boosted throughput to
> 25k/s from 13k/s.
> 
> Bypassing the json.loads bumped it to ~60k/s.
> 
> cProfile + snakeviz shows 5% is taken by `isinstance()` so maybe it's
> worth generating loop body code ahead of time based on column types.
> 
> Defaults to use Python psycopg2+ujson for TSJ as it shows better
> performance than taking the tsv2tsj route.


3. Rewrites db-query-tsj to codegen Python ahead of time (a8a4811c42fb2d39d2692a0511391725d5ac5a7f)
> Since every row has fixed number of columns within a single execution,
> use the type info to generate code that uses direct accesses with
> indexes.  No need to pay the cost of iterating over collections
> dynamically.
> 
> This gives 1.5x increase in throughput: 60k/s to 95k/s
> and now with parallel unloads (e.g., `deepdive unload sentences
> s-{0..7}.tsj` it can go up to 200MiB/s on the latest MacBook Pro)

File: database/db-driver/postgresql/db-query-tsj (hunk 2 above)
Comment (netj) [inline database/db-driver/postgresql/db-query-tsj]: Thanks for the pointers. I was referring to `entry_points`.  Maybe I'll have to keep the script with the comments for `usage` and a bit of sys.argv mapping for consistency within DD.  I don't think having a main that does little is very important for the library.

Comment (netj) [inline database/db-driver/postgresql/db-query-tsj]: On second thought I think we should defer this to when we have at least two more concrete use cases.  While attempting, I realized not many Python code will want to generate TSJ files directly from database as it can just delegate to this executable, and to make any smaller unit useful by itself, we have to generalize and add significant bloat to this highly specialized code.  I can't easily figure a way to expose meaningful knobs but still keep it performant.  Currently the SLOC < 50, but I think making it testable, reusable at finer granularity will just make it 2x or even longer, making it ironically less maintainable.  The executable unit is already tested fairly extensively from `deepdive_sql.bats`, and I don't think we need tests for every few lines of the code.  Maybe you can take a stab later when some other code needs to output TSJ after executing a SQL query.

File: ddlib/ddlib/util.py (hunk 4 above)
Comment (netj) [inline ddlib/ddlib/util.py]: You mean `parse_json`?  It's just a `json.loads` with context-dependent error message translation.. that I doubt would be reusable or worth testing.  It's just to make the for loop body look cleaner with list comprehension syntax.  If the nested `def`s disturbs you in general, I can inline the code and make the following for loop longer.  I also hate `def`s but Python's dichotomy btwn statements and expressions forces you to use them.

File: database/db-driver/postgresql/db-query-tsj [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+#psycopg2.extras.register_default_json(loads=ujson.loads)
+
+# use pyscopg2 to access the database
+conn = psycopg2.connect(
+        # NOTE these environment variables are supposed to set by db-parse
+        database=os.environ.get("DBNAME"),
+        password=os.environ.get("DBPASSWORD"),
+        user=os.environ.get("DBUSER"),
+        host=os.environ.get("DBHOST"),
+        port=os.environ.get("DBPORT"),
```
alldefector: We currently use the DBOPTIONS env var (`search_path=abc,public`) to achieve specify which schema DD should use. That's not a good design but has been used in many places. Can we make this pick that up?

netj: Sure. Is it `PGOPTIONS` or `DBOPTIONS`?  I think the underlying libpq is supposed to pick up a bunch of env vars and psycopg2 also claims that but it didn't work, hence these lines.

alldefector: Sorry, it's `PGOPTIONS`...

shahin:
@netj Thanks for pasting those numbers. I remember seeing them before, but they don't contain any comparison to JSON lines. Since JSON solves our NULL-encoding problems and is very common, seems to me like this is the proper baseline for comparison.

I agree that we can cook up any performance numbers we want by picking and choosing tables. I'm looking for persuasive numbers to convince new hires that they should learn and maintain this new format instead of JSON lines.

Maybe we have a bunch of really wide tables with really long keys. Or Deepdive is peppered with cases where parsing with `jq` instead of `cut` is way too slow. I'm sure concrete examples of this seem obvious to those deeply familiar w/ Deepdive code! I guess I'm hoping we can dig up two or three of them and paste them here.

Put another way, seems like we'd make these claims about JSON lines when making a case for TSJ (slightly modified from your TSJ version above):

1. JSON lines parsing (decoding/deserialization) is super easy: ~~split each input line by TAB, then parse each field~~ parse each line with universally available JSON parser.
2. JSON lines formatting (encoding/serialization) is super easy: encode each output ~~value~~ line in JSON, again universally available~~, then join with TAB to produce a line.~~
3. Extra JSON *lines* encoding overhead is ~~negligible~~ *not negligible* compared to TSV.

Since 1 and 2 are clear wins for JSON lines, our argument for TSJ seems to rest on 3. Can we show that JSON encoding overhead is not just positive but also *non-negligible* for our use cases, compared to TSV? Compared to TSJ?

netj:
Alright, here're some numbers.  Still using the sentences-1000 data.

## Size
```
$ wc s.*
   26975   836437 31206139 s.tsj
   26975   836434 31091937 s.tsv
   26975   593662 34524064 s.jsonl
```

All ~27k rows and ~30MiB.  tsj is 1.003x the size of tsv and jsonl is 1.11x.
Interestingly jsonl has the least number of word count, which suggests it uses almost no space, and it must be in the most compact form.

## Throughput
```
$ datasource() { local ext=$1; echo >&2 "# Using s.$ext"; cat $(for i in $(seq 1000); do echo s.$ext; done); }
$ datasink() { timeout 10s cat >/dev/null; }
```

```
$ datasource tsj | pv --bytes --timer --average-rate --cursor | datasink
# Using s.tsj
10.4GiB 0:00:10 [1.04GiB/s]
$ datasource tsv | pv --bytes --timer --average-rate --cursor | datasink
# Using s.tsv
10.4GiB 0:00:09 [1.05GiB/s]
$ datasource jsonl | pv --bytes --timer --average-rate --cursor | datasink
# Using s.jsonl
10.4GiB 0:00:10 [1.04GiB/s]
```

$ datasource tsj | pv --bytes --line-mode --timer --average-rate --cursor | datasink
# Using s.tsj
5.76M 0:00:09 [ 576k/s]
$ datasource tsv | pv --bytes --line-mode --timer --average-rate --cursor | datasink
# Using s.tsv
5.94M 0:00:09 [ 594k/s]
$ datasource jsonl | pv --bytes --line-mode --timer --average-rate --cursor | datasink
# Using s.jsonl
5.15M 0:00:09 [ 515k/s]

Projection

$ datasource tsj | pv --bytes --line-mode --timer --average-rate --cursor | cut -f2,8 | datasink
# Using s.tsj
3.94M 0:00:10 [ 394k/s]
$ datasource tsv | pv --bytes --line-mode --timer --average-rate --cursor | cut -f2,8 | datasink
# Using s.tsv
3.99M 0:00:09 [ 399k/s]
$ datasource jsonl | pv --bytes --line-mode --timer --average-rate --cursor | jq -c '{sentence_index:.sentence_index, doc_offsets:.doc_offsets}' | datasink
# Using s.jsonl
 187k 0:00:09 [18.7k/s]

$ datasource tsj | pv --bytes --line-mode --timer --average-rate --cursor | cut -f1 | datasink
# Using s.tsj
4.13M 0:00:10 [ 412k/s]
$ datasource tsv | pv --bytes --line-mode --timer --average-rate --cursor | cut -f1 | datasink
# Using s.tsv
4.11M 0:00:09 [ 411k/s]
$ datasource jsonl | pv --bytes --line-mode --timer --average-rate --cursor | jq -c '{doc_id:.doc_id}' | datasink
# Using s.jsonl
 210k 0:00:10 [21.1k/s]

$ datasource tsj | pv --bytes --line-mode --timer --average-rate --cursor | cut -f2 | datasink
# Using s.tsj
4.19M 0:00:09 [ 419k/s]
$ datasource tsv | pv --bytes --line-mode --timer --average-rate --cursor | cut -f2 | datasink
# Using s.tsv
4.23M 0:00:10 [ 422k/s]

$ datasource jsonl | pv --bytes --line-mode --timer --average-rate --cursor | jq -c '{sentence_index:.sentence_index}' | datasink
# Using s.jsonl
 217k 0:00:10 [21.7k/s]
$ datasource jsonl | pv --bytes --line-mode --timer --average-rate --cursor | jq -c '{x:.sentence_index}' | datasink
# Using s.jsonl
 223k 0:00:10 [22.2k/s]
$ datasource jsonl | pv --bytes --line-mode --timer --average-rate --cursor | jq -c '[.sentence_index]' | datasink
# Using s.jsonl
 224k 0:00:10 [22.4k/s]
$ datasource jsonl | pv --bytes --line-mode --timer --average-rate --cursor | jq -c '.sentence_index' | datasink
# Using s.jsonl
 227k 0:00:10 [22.7k/s]
```

netj:
Alright, here're some numbers, in a notebook!  (I wish I was more fluent to produce nice charts as well, but that's for next time :)

https://gist.github.com/netj/c5ef35e74488500da95701f90e471765

Hope you are convinced why I think JSON lines (or jsonl) doesn't make sense but TSJ does.  I still have more cooking to do to make TSJ appear better than TSV, but I hope you get the idea.

shahin:
Very convincing notebook, thanks @netj . Looks like ad-hoc inspection is the main use case. Other than what's here, do you have any bottlenecks in mind that we need to migrate to TSJ?

Otherwise LGTM.

PR #568, opened 2016-07-23
Fix deepdive.calibration.observation_query

alldefector:
... though I don't understand how "observation" vars are used...

### DIFF
--- a/compiler/compile-config/compile-config-2.01-grounding
+++ b/compiler/compile-config/compile-config-2.01-grounding
@@ -199,8 +199,8 @@ def factorWeightDescriptionSqlExpr:
             "
           end
         , if $deepdive.calibration.observation_query then
-            # run user holdout query if configured
-            "\($deepdive.calibration.holdout_query)"
+            # run user observation query if configured
+            "\($deepdive.calibration.observation_query)"
           else empty
           end
         ] | map("deepdive sql \(asPrettySqlArg)") | join("\n"))


netj:
I think @thodrek convinced @feiranwang to add this.  Could you give a brief motivation for observation vars?

thodrek:
@netj : Those where already there. Not sure about the initial motivation. The reason I was asking for those was to be able to introduce observed variables in a factor graph and have a single factor outputting different values based on the value of those observed variables. At that point factors in DD didn't support constants as input.

alldefector:
The only place where `variable.is_observation` is used seems to be only at the beginning of `GibbsSamplerThread::sample_sgd_single_variable`: `if (variable.is_observation) return;` Not sure if there is any code for "outputting different values based on the value of those observed variables"... (perhaps quasi-hard rules in ddlog?)

Regardless, looks like this bug has been there since Dec 2015 (when netj revamped grounding with jq) -- and nobody has noticed! Assuming that `is_evidence` works for the intended use case, perhaps let's just purge `is_observation`?

netj:
I think their main value is providing a set of constants for factors that don't participate in training, which cannot be done with evidence variables alone.  I'm okay with either way: keeping with the fix or purging as nobody seems to actually care that much.

alldefector:
OK, to minimize effort, I'm landing this for now. Next time we make any related changes / refactoring, we can do the merciless purge...

Issue #569, opened 2016-07-23
Bug: Incomplete loading of domains from grounding


alldefector:
We have a sparse categorical app where the inference result (`dd_inference_result_variables`) contain probs for cid = 0,1,2,3,... while those probs should have been for cid = 6,44,333,555,...

After some debugging, I found that the broken step appears to be in `--domains <(flatten factorgraph/domains)` of `process/model/learning`.

The domain dump SQL is correct, covering all categorical vars. The dumped binary files (`domains.part-*.bin.bz2`, we have 31 parts) also appear to be correct -- their total byte count matches the expected size (`select sum(8 + 4 + 4 * dd__count) from dd_variables` == `find -L \"$@\" -type f -size +0 -exec pbzip2 -c -d -k {} + | wc --bytes`). However, the var count from `FactorGraph::load_domains` does not match the total var count (150K or so out of 180K or so; and the ratio is different from run to run).

So, it appears that something is wrong with the `flatten` pipe into the sampler...

netj:
See if grounding with the following produces correct result:
```
export DEEPDIVE_NUM_PROCESSES=1
```
You can probably bunzip2 the single domain binary and feed the sampler with the rest of inputs to see if flatten is the culprit or something else.

I'm suspecting the `load_domains` may be terminating earlier..

alldefector:
OK, that's indeed related. Here is what I tried:
```
rm -f run/model/grounding/variable/link/domains.part*
DEEPDIVE_NUM_PROCESSES=XXX deepdive redo process/grounding/variable/link/dump_domains process/grounding/combine_factorgraph process/model/learning
```

I tried different values of XXX. If XXX is 1 or 20, var count from `load_domains` is correct. If XXX is 28 or above, var count from `load_domains` becomes smaller than expected.

If I just run
```
DEEPDIVE_NUM_PROCESSES=30 deepdive redo process/grounding/combine_factorgraph process/model/learning
```
the var count is always correct.

So, something seems fishy with the concurrency control of [`deepdive compute execute`](https://github.com/HazyResearch/deepdive/blob/master/runner/compute-driver/local/compute-execute) or perhaps file flushing in `sampler-dw text2bin domain /dev/stdin >(pbzip2 >domains.part-${DEEPDIVE_CURRENT_PROCESS_INDEX}.bin.bz2)`, or the symlink creation in `combine_factorgraph`... (In other words, `load_domains` tried to read the part files before they are ready.)

[...]

PR #570, opened 2016-07-25
Fix dump_domains race condition

alldefector:
Partial fix for https://github.com/HazyResearch/deepdive/issues/569

Don't have a good solution for vars and factors dumping yet.


netj:
LGTM

PR #573, opened 2016-07-31
Add conda recipe and supporting fixes

shahin:
This change adds a conda recipe that (1) fetches dependencies, (2) builds a binary deepdive package, and (3) configures a conda environment with deepdive's required environment variables.

Benefits of conda packaging include:

A. Installation of deepdive and dependencies without sudo or root privileges,
B. Fast (binary) installs,
C. Can declare deepdive as a dependency in other conda projects,
D. Sandboxed deepdive instances inside conda environments,
E. Eventual removal of custom bundled-dependency code, and
F. Eventual removal of brittle, manually-managed PATH, PYTHONPATH, [...] environment variables

Supporting fixes on this branch include:

a. Making bundled dependencies optional,
b. Python 3 compatibility fixes, and
c. Fixes to the spouse example

Note that not all dependency packages are available on public repositories, and we can't turn off bundled dependencies or turn on automatic conda packaging for deepdive until they are.

### DIFF
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -1,0 +1,21 @@
+if [ "$(uname)" == "Darwin" ]; then
+  # C++11 finagling for Mac OSX
+  export CC=clang
+  export CXX=clang++
+  export MACOSX_VERSION_MIN="10.9"
+  CXXFLAGS="${CXXFLAGS} -mmacosx-version-min=${MACOSX_VERSION_MIN}"
+  CXXFLAGS="${CXXFLAGS} -Wno-error=unused-command-line-argument" 
+  export LDFLAGS="${LDFLAGS} -mmacosx-version-min=${MACOSX_VERSION_MIN}"
+  export LINKFLAGS="${LDFLAGS}"
+  export MACOSX_DEPLOYMENT_TARGET=10.9
+
+  # make sure clang-format is installed
+  # e.g. brew install clang-format
+
+  # make sure autoreconf can be found
+  # e.g. brew install autoconf
+
+  # for graphviz, also brew install autogen libtool
+fi
+
+make install PREFIX=$PREFIX
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -26,7 +26,7 @@ cat << EOF > ${PKG_NAME}-env-activate.sh
 #!/usr/bin/env bash
 
 export PRE_${PKG_NAME}_PATH=\$PATH
-export PATH=\$CONDA_PREFIX/util:$PATH
+export PATH=\$CONDA_PREFIX/util:\$PATH
 EOF
 
 cat << EOF > ${PKG_NAME}-env-deactivate.sh
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -19,3 +19,25 @@ if [ "$(uname)" == "Darwin" ]; then
 fi
 
 make install PREFIX=$PREFIX
+
+# add the /util directory to the PATH inside this conda env
+# http://conda.pydata.org/docs/using/envs.html#saved-environment-variables
+cat << EOF > ${PKG_NAME}-env-activate.sh
+#!/usr/bin/env bash
+
+export PRE_${PKG_NAME}_PATH=\$PATH
+export PATH=\$CONDA_PREFIX/util:$PATH
+EOF
+
+cat << EOF > ${PKG_NAME}-env-deactivate.sh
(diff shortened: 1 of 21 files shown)


alldefector:
LGTM!

shell/jq
```diff
@@ -1,8 +1,8 @@
-#!/bin/sh
+#!/bin/bash
```
netj: Please keep this `/bin/sh`.  Maybe `type -p -a` below needs to be `which -a`..

conda.recipe/build.sh
```diff
@@ -0,0 +1,43 @@
...
+
+  # make sure autoreconf can be found
+  # e.g. brew install autoconf
+
+  # for graphviz, also brew install autogen libtool
+fi
+
+make install PREFIX=$PREFIX
+
+# add the /util directory to the PATH inside this conda env
```
netj: Not sure what a conda env is, but why do we want to expose everything in util/?  Most of them aren't standalone executables, i.e., won't work outside an env set up by bin/deepdive.. so this will introduce a lot of broken commands to the conda env.

ddlib/ddlib/util.py
```diff
@@ -98,7 +98,7 @@ def normalize_type_name(ty):
   if ty in CANONICAL_TYPE_BY_NAME:
     return CANONICAL_TYPE_BY_NAME[ty]
   else:
-    for patt,ty_canonical in CANONICAL_TYPE_BY_REGEX.iteritems():
+    for patt,ty_canonical in CANONICAL_TYPE_BY_REGEX.items():
```
netj: Seems like we're sacrificing Python2 performance for code compatibility with Python3.. Is there a way to still use generators in py2 for simple iterations?  Or is there proof that the difference is negligible?

netj:
@shahin Thanks for the fixes and conda packaging.  In addition to my inline suggestions, could you add a simple note on maintaining the conda package in doc/developer.md?

Most benefits of conda you listed sounded like tautology to me, i.e., "conda packaging is good for using conda," not solving any new problem, but the first reason is good enough for a complete switch: "A. Installation of dependencies without sudo."

Also, please keep in mind that DD has to allow two different PYTHONPATHs (or environments in general): user/app's and deepdive's own.  So far, the distinction has not been clear throughout DD's code, often causing strange UDF issues or DD crashes.  Naively forcing all DD users to share a single PYTHONPATH and PATH with conda will make certain UDFs impossible to use with DD.  Greenplum is a good example of such nightmare as users are misguided to switch to their internal python for accessing the database.

ddlib/ddlib/util.py
shahin: This dictionary has only six things in it and it's a constant, so the risk of impacting performance is small and could go in either direction. We can definitely find some way to use an iterator in both py2 and py3 here, but it'll make the code more complex. In your experience, is this piece of code a hot path?

shahin:
> Most benefits of conda you listed sounded like tautology to me, i.e., "conda packaging is good for using conda,"

Ha ha, I did use the word "conda" there many times. It probably reads tautologically because the word is semantically ambiguous. "conda" refers to any subset of:

1. A package format,
2. A dependency system,
3. An environment manager.

On ubuntu you have deb/apt/nothing (or chroot?), in Python you have setup.py/pip/virtualenv, in Rust you have crate/cargo/multirust. The package format in this PR buys us a ticket to the dependency system (in C) and the environment manager (in D) the same way that Python packaging gets you pip and virtualenv. The conda/conda/conda naming situation we have here is pretty unfortunate.

> Also, please keep in mind that DD has to allow two different PYTHONPATHs (or environments in general): user/app's and deepdive's own [...] Naively forcing all DD users to share a single PYTHONPATH and PATH with conda will make certain UDFs impossible to use with DD. Greenplum is a good example of such nightmare as users are misguided to switch to their internal python for accessing the database.

This isn't quite clear to me yet.

Greenplum is a process that runs as a service. So on the one hand it's entitled to its own runtime environment, but on the other it's a huge pain to write and debug Greenplum UDFs because they depend on this special runtime environment that interacts with user environments in brittle ways. I think this is what your last sentence is referring to.

Ideally, Greenplum would be installed in its own conda environment that can be activated and deactivated anywhere by anyone (I have no intention of doing this atm). UDFs would be written, tested, and deployed inside this conda environment.

It sounds like you're saying the Deepdive situation is similar to Greenplum's. Is that right? I'd be grateful for more details or a pointer to a good example UDF.

alldefector:
@shahin could you merge or rebase master?

PR #584, opened 2016-08-23
Always analyze relation after deepdive-load

zifeishan:


database/deepdive-load
```diff
@@ -180,5 +180,8 @@ done
 # load everything queued so far
 loadQueued
 
+# always analyze relation after loading
+deepdive db analyze "$Relation"
```
alldefector: Might be better to do `deepdive-db analyze` given the `deepdive-mark done` pattern?

PR #594, opened 2016-09-18
Version conda builds using git

shahin:
We'd like to build conda packages without having to specify the version number manually. The version of the conda package should always correspond to the version according to git.

Use git to version the conda package, so that:

1. The most recent tag on the branch being built is used as the version number, and
2. The number of commits since that tag on this branch is used as the build number.

(2) means that we need to stop rebasing non-local branches. Instead, use `git merge`.

Note that the most recent tag on `master` is `UNSTABLE` which isn't a version number:

```
$ git describe --abbrev=0 --tag
UNSTABLE
```

Once this is merged in, we should tag the merge commit with a new version (v0.8.1?) so that the resulting package has a name that makes sense.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: deepdive
-  version: "0.8"
+  version: {{ GIT_DESCRIBE_TAG|replace("v","") }}
 
 source:
   git_url: ../
@@ -8,7 +8,7 @@ source:
 build:
   script_env:
     - USER
-  number: 1
+  number: {{ GIT_DESCRIBE_NUMBER|int }}
 
 test:
   files:


alldefector:
@netj looks like dockerhub access is flaky:
```
$ DOCKER_IMAGE_TO_PUSH=$DOCKER_IMAGE_TEST_PREFIX; if [[ $TRAVIS_TEST_RESULT = 0 ]]; then DOCKER_IMAGE_TO_PUSH+=PASS; else DOCKER_IMAGE_TO_PUSH+=FAIL; fi;
after_script.2
0.01s$ if [[ $TRAVIS_PULL_REQUEST = false ]]; then tag=travis.$TRAVIS_BRANCH; else tag=travis-pr$TRAVIS_PULL_REQUEST; fi; tag=$(echo -n "$tag" | tr -c 'A-Za-z0-9_.-' _);
$ docker login -e "$encrypted_DOCKER_EMAIL" -u "$encrypted_DOCKER_USERNAME" -p "$encrypted_DOCKER_PASSWORD";
Flag --email has been deprecated, will be removed in 1.13.
Login with your Docker ID to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com to create one.
Username: 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

netj:
@alldefector Looks like the credentials are not decrypted for non-HazyResearch github users.  Will fix travis.yml to skip the pushing instead of hanging

PR #602, opened 2016-10-26
Use View for factor tables in Grounding

zifeishan:
In many applications, factor tables are huge tables unnested from compact original representations. This change to use views instead of materializing them is measured to save 30--40% of overall grounding time in some application.


zifeishan:
Breaking spouse example grounding, discarding. In the future may want to make this dynamic / configurable.

PR #615, opened 2017-01-01
Make gpfdist work

alldefector:
Fixes two issues:
1. An unload query with ambiguous columns (including duplicate `cid`s and `?column?`s) would break gpfdist.
2. `deepdive compute execute` hangs indefinitely if spawned `gpfdist` command fails.

### DIFF
--- a/database/db-driver/greenplum/db-unload
+++ b/database/db-driver/greenplum/db-unload
@@ -56,6 +56,7 @@ case $format in
             }
             trap cleanup EXIT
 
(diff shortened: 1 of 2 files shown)


database/db-driver/greenplum/db-unload
```diff
@@ -56,6 +56,7 @@ case $format in
             }
             trap cleanup EXIT
 
+            gpfdist --help &>/dev/null ||  { echo 'gpfdist not working! make it work!'; exit 1; }
```
netj: `error` command can be used in these cases.

PR #617, opened 2017-01-07
Added --bignum flag to awk to restore vid

zifeishan:


compiler/compile-config/compile-config-2.02-learning_inference
```diff
@@ -191,8 +191,8 @@ if .deepdive_.execution.processes | has("process/grounding/combine_factorgraph")
             for pid in $DEEPDIVE_FACTORGRAPH_SHARDS; do
                 cd \"$DEEPDIVE_RESULTS_DIR\"/$pid
                 cat inference_result.out.text |
-                # restoring shard ID to the vids
-                awk -v SHARD_BASE=$(($pid << 48)) '
+                # restoring shard ID to the vids. Note that awk --bignum flag is required for precision
+                awk --bignum -v SHARD_BASE=$(($pid << 48)) '
```
netj: This flag seems to be something relatively new?  I can't find it from my Mac or Linux's gawk, mawk, or BSD awk. AWK has too many forks and versions out there, so maybe we should use something simpler.  I'll find a better solution and fill it in here.  Thanks for suggesting the fix and hope this one works for you for now.

PR #621, opened 2017-01-12
Fix jq load_probabilities precision error using python

zifeishan:
`jq` has a precision bug while loading probabilities with big vids, due to a known issue that it's using double floating numbers to store integers. There seems no good way in jq to support better precision:
https://github.com/stedolan/jq/issues/369

We've tried using awk with --bignum but versioning seems a problem.

Here we just try to use python3 to do the job.

### DIFF
--- a/compiler/compile-config/compile-config-2.02-learning_inference
+++ b/compiler/compile-config/compile-config-2.02-learning_inference
@@ -197,10 +197,11 @@ if .deepdive_.execution.processes | has("process/grounding/combine_factorgraph")
                 cd \"$DEEPDIVE_RESULTS_DIR\"/$pid
                 cat inference_result.out.text |
                 # restoring shard ID to the vids
-                jq -R -r --argjson SHARD_BASE $(($pid << 48)) '
-                    split(\" \") |
-                    [ (.[0] | tonumber + $SHARD_BASE | tostring)
-                    , .[1:][] ] | join(\"\\t\")' |
+                python3 -c "
+import sys
+for line in sys.stdin:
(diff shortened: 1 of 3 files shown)


compiler/compile-config/compile-config-2.02-learning_inference
```diff
@@ -197,10 +197,11 @@ if .deepdive_.execution.processes | has("process/grounding/combine_factorgraph")
                 cd \"$DEEPDIVE_RESULTS_DIR\"/$pid
                 cat inference_result.out.text |
                 # restoring shard ID to the vids
-                jq -R -r --argjson SHARD_BASE $(($pid << 48)) '
-                    split(\" \") |
-                    [ (.[0] | tonumber + $SHARD_BASE | tostring)
-                    , .[1:][] ] | join(\"\\t\")' |
+                python3 -c "
```
alldefector: I think you need to escape the double quote (with backslash). Travis is complaining about jq compilation.

Also, @netj should we just use `python` here -- which should be compatible with both 2 and 3.

@zifeishan let's do some perf test for this?

netj: I couldn't resist doing a [benchmark notebook](https://gist.github.com/netj/20e00e1719684ce08f38edff519f575a)
and the winner is.... Perl!

jq was actually the slowest, ~7x slower than Perl and 3x than AWK.  Python is ~1.7x, AWK is ~2.2x.
jq and AWK uses little memory ~10MB, ~1/3 of Python (35-45MB), but so does Perl (14MB).
So Perl is a clear winner.

alldefector: If perl can handle all 2^64 numbers and that's not subject to version differences on different OSes, we can go with perl.

PR #622, opened 2017-01-15
Fix hanging when unload fails

alldefector:
DD hangs if gpfdist fails (in unload background process).

### DIFF
--- a/database/db-driver/greenplum/db-unload
+++ b/database/db-driver/greenplum/db-unload
@@ -6,7 +6,7 @@
 # parallel connections, which MAY RESULT IN INCORRECT RESULT UNLESS QUERY IS
 # TOTALLY ORDERED BY COLUMNS WITH NO DUPLICATE VALUES!
 #
-# GPFDIST_PORT_BASE (defaults to 10000) is the base port number for the gpfdist
+# GPFDIST_PORT_BASE (defaults to 12000) is the base port number for the gpfdist
 # processes.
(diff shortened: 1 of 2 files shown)


database/db-driver/greenplum/db-unload
```diff
@@ -6,15 +6,15 @@
...
+# GPFDIST_PORT_BASE (defaults to 12000) is the base port number for the gpfdist
 # processes.
 # GPFDIST_MAX_LENGTH (defaults to 1MiB) may need to be increased for larger
 # records to be unloaded through gpfdist.
 # GPFDIST_LOG_PATH can be set to a path prefix for logging.
 ##
 set -eu
 
-: ${GPFDIST_DISABLE:=} ${GPFDIST_PORT_BASE:=10000} ${GPFDIST_MAX_LENGTH:=$((2**20))} ${GPFDIST_LOG_PATH:=}
+: ${GPFDIST_DISABLE:=} ${GPFDIST_PORT_BASE:=12000} ${GPFDIST_MAX_LENGTH:=$((2**20))} ${GPFDIST_LOG_PATH:=}
```
netj: While we're at this let's also randomize the base?
`$(( 100*($RANDOM % 500 + 100) ))` which would give us 10000 through 59900

alldefector: We should make it pick ports based on availability. I moved it to 12000 for now because I used 10000 somewhere else...

alldefector: Closing for now. We can change that later.

